PR #8, opened 2015-07-24
add feature to check in models into repo using REST

erlemantos:


aedelmann:
I saw that your REST upload handles the exception and sends back the http error code. I would suggest to use the REST Template that takes over this kind of stuff :) Check out some other REST methods using the template for your reference.

aedelmann:
You are using jersey multipart library. I am going to create the needed CQ for this.

erlemantos:
Now using restTemplate. See commit on Jul 27, 2015.

PR #10, opened 2015-07-29
For Bug 473708 - REST implementation of Repository API 

erlemantos:
This is for Bug 473708 [ https://bugs.eclipse.org/bugs/show_bug.cgi?id=473708 ]
- First checkin of repository bundle
- Renamed some files that are previously interfaces but are now POJO


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=473708

erlemantos:
Addendum to pull request: 

1. Added an extension point for model repository in vorto.core bundle. 
2. Added an extension for the model repository using the REST api in vorto.repository bundle.

aedelmann:
In this case I would introduce a IModelSerializer which accepts a Model object and returns a String Object. There are then 2 implementations, one to serialze it to XMI, another one to the DSL. 
Btw: The implementations would use the existing EMF resource and Xtext Resource respectively.

What do you think?

bundles/org.eclipse.vorto.core/src/org/eclipse/vorto/core/api/repository/ModelContent.java
```diff
@@ -23,30 +23,52 @@
  *	This class represents a resource that we can get from
  *  the repository. 
  */
-public interface IModelContent {
+public class ModelContent {
+	
+	private byte[] content;
```
aedelmann: Model Content should not expose the content, because then the using component would need to know in what format it is. I would recommend to make the ModelContent and thus IModelRepository just be aware of Model objects and make it technology independent (Inpustream)

erlemantos:
> What do you think?

Yes, definitely a good way to refactor it. An IModelSerializer, two implementers and a factory to decide which serializer to return.

erlemantos:
Ok, how about this --

We create an interface called IModelSerializer with three methods --

- Boolean canHandle(String operation)
- String serialize(Model model) 
- Model deserialize(String representation)

I will then make an extension point for this IModelSerializer, because the
two implementations will need to be in different bundles. The XMI to Model serializer
can be implemented in vorto.core since it doesn't need Xtext, but the DSL to Model serializer
needs to be implemented in vorto.editor since it needs Xtext.

I will then create a factory that takes an operation string and returns an IModelSerializer based on the operation string. This factory has to go thru ALL the extensions and see which among them can handle the operation.

- IModelSerializer getModelSerializer(String operation)

erlemantos:
Moved the XMI conversion code to vorto.repository to avoid polluting the vorto.core for now. This issue should be revisited later.

aedelmann:
A model serializer exposing a deserialize functionality kinda clashes. I would separate this behind its own interfaces.

erlemantos:
The model serializer concept has already been removed in the current state of the code.

aedelmann:
Looks good!

PR #13, opened 2015-08-31
Model Repository Integration

erlemantos:
Changes-

- Shared models that are already on the local workspace are not downloaded anymore.
- URL for downloading models are constructed using the correct model name, not the lowercase version of it.
- Usage of instanceof instead of isInstance for situations where the class we want to check is already known at compile time.


aedelmann:
Cool. Thanks.

PR #21, opened 2015-09-04
Changed "Check-in to repository" to "Share to repository"

erlemantos:


erlemantos:
I've been using that same sign-off ever since I started committing. I don't understand why suddenly now it becomes invalid. Can any eclipse guys look at this?

PR #23, opened 2015-09-04
Changed the context menu for checking-in. From "Check-in to repository" to just "Share"

erlemantos:


erlemantos:
Ok, this requires intervention from eclipse folks. I don't understand why my sign-off isn't working. I've even changed the casing (to all lowercase) of my email address.

PR #26, opened 2015-09-21
Changed "Check-in to repository" to "Share" and placed all Vorto preferences under one root preference item. 

erlemantos:
- Changed "Check-in to repository" to "Share"
- Placed all Vorto preferences under one root preference item.


erlemantos:
Pull request still not pushing through due to Eclipse bug.

PR #31, opened 2015-10-23
ignore files generated by eclipse

alex-ellwein:
This is about to ignore files generated by eclipse and 
remove existing generated files.

Signed-off-by: Alexander Ellwein <alexander.ellwein@bosch-si.com>


aedelmann:
Great, Alex, thank you!

PR #37, opened 2016-02-02
Generated visual representation of the VORTO meta model

ChristophRathfelder:
As proposed by Olaf, I created a PullRequest for the visual representation of the meta model


OlafWeinmann:
Hi Christoph,

thank you very much! It seems to me that your sign-off did not work. Did you sign a CLA with Eclipse? If not, could you please do so? 
Would it be possible to remove the png from the pull request since it can be generated?

Thanks in advance!

Olaf

PR #39, opened 2016-02-03
Generated Visual Representation of Vorto MetaModel

ChristophRathfelder:
Using the ecore files as input, I generated a visual representation of the VORTO Metamodel


ChristophRathfelder:
Added the signed-off-footer afterwards to pass the IP check.

PR #41, opened 2016-02-23
DSL Improvements for Property

erlemantos:
Set measurementUnit to an enumeration literal type so you can do something like this:

```
using com.mycompany.type.Temperature ; 1.0.0
...
mandatory heat as int with { measurementUnit : Temperature.C }

```

Set readable, writable, and eventable to Boolean type.

`mandatory heat as int with { eventable: true, writable : false, readable : false }`


aedelmann:
Hi Erle, that's a nice feature for the DSL. Thanks!

PR #44, opened 2016-03-01
App properties changes for eclipse environment

nagavijays:
Default context path changed for eclipse environment.


aedelmann:
Hi Vijay, 
please squash your commit with merge --squash and then sign off only that single commit with commit -s

Cheers ,Alex

PR #45, opened 2016-03-02
Fixes Bug 487483 - Mapping-rules of events are ignored when generating code 

erlemantos:
- Fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=487483
- Pass local mapping context to generators when generating from the vorto toolset


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=487483

Issue #63, opened 2016-03-14
Mapping-rules of events are ignored when generating code


nagavijays:
Migrated from https://bugs.eclipse.org/bugs/show_bug.cgi?id=487483. 
pull request: #45

erlemantos:
Fix for this issue currently in pull request.

erlemantos:
Pull request has been merged. We can resolve this issue now.

PR #116, opened 2016-03-21
Fix for https://github.com/eclipse/vorto/issues/63 - Mapping-rules of events ignored

erlemantos:
Fix for https://github.com/eclipse/vorto/issues/63 - 
Mapping-rules of events are ignored when generating code.


aedelmann:
Looks good to me :)

PR #147, opened 2016-04-20
Bulk upload models with review points.

### LINKED ISSUE #68: Bulk upload of models (zip archived)
Migrated from https://bugs.eclipse.org/bugs/show_bug.cgi?id=488609

nagavijays:
Bulk upload models with review points implemented.
Closes #68


server/repo/repo-core/src/main/java/org/eclipse/vorto/repository/service/IModelRepository.java
```diff
@@ -67,6 +68,15 @@
 	UploadModelResult upload(byte[] content, String fileName);
 	
 	/**
+	 * Uploads multiple model files in zip and validates it.
+	 * If all the models are valid then the models are checked in using array of {@code} ModelHandle 
+	 * {@link IModelRepository#checkinMultiple(ModelHandle)}}
+	 * @param fileName
+	 * @return result about information of the uploaded content and the upload handle. 
+	 */
+	List<UploadModelResult> uploadMultipleModels(String fileName);
```
aedelmann: Why can't we re-use the existing upload(byte[] content, String fileName) ? 
I would assume that the method implementation could use 2 different handle strategies based on the fileName type ,one for zip with multiple files and one for a single model file.

server/repo/repo-core/src/main/java/org/eclipse/vorto/repository/service/IModelRepository.java
```diff
@@ -78,6 +88,17 @@
...
 	/**
+	 * @pre All Models are validated for integrated checks. 
+	 * 
+	 * @post Models are stored into repository and sends notifications to watchers. 
+	 * 
+	 * Checks in multiple models into the repository.
+	 * @param uploadHandle - all model file upload with file handles.
+	 * @param author
+	 */
+	void checkinMultiple(ModelHandle[] modelHandles, String author);
```
aedelmann: Same as previous comment. Please use the existing checkin(string handle, string author) instead of introducing another interface method doing a similar thing.

server/repo/repo-core/src/main/java/org/eclipse/vorto/repository/service/IModelRepository.java
nagavijays: Yes. The existing single file upload validations criteria are different which is not suitable for multiple model upload and also to avoid multiple conditional checks for single and multiple upload files.

PR #232, opened 2016-06-07
Added runtime test module for Information model

shiv12095:
This will generate the files for the runtime test module for information model. This was missed out in https://github.com/eclipse/vorto/pull/227


aedelmann:
Hi realized, that your previous pull request did not migrate functionblock, datatype and mapping mwe workflow to xtext 2.9.2 syntax. That might also be a reason for the build to fail. Could you also please correct that?

shiv12095:
I am able to build the functionblock, datatype and mapping using xtext 2.9.2 without actually changing the workflow. I have tried that using the Xtext plugin available on Eclipse and using maven to build each of these individually. They throw no errors. I am still not able to replicate the problem that occurred on Hudson

Can you merge this PR and try building it again? Maybe that might solve it. If not, I will migrate the others according to the new 2.9.2 syntax

PR #238, opened 2016-06-10
Adjusting default context path for eclipse environment.

nagavijays:
Adjusting default context path for eclipse environment. Also removed hardcoded user during startup.
Signed-off-by: Nagavijay Sivakumar <nagavijay.sivakumar@bosch-si.com>


aedelmann:
There are conflicts , please resolve them and Pull Request again. 
Thanks!

nagavijays:
There are no conflicts now. Hope it should it can be merged now.

aedelmann:
Hi Vijay, 
thanks for straightening out the merge problem. 
However I can see that you are using the same commit message and it would appear twice in the history which is rather not so nice. In the future please use --amend to add amend to the previous commit. 
For the current case , can you squash your commits into a single commit and pull request again?

Thanks!

Cheers, Alex

nagavijays:
Hi Alex, 
Squashed into single commit. Earlier squash didn't work. I have to use different method to merge. I think it looks good now.

PR #292, opened 2016-06-23
Closes #222 - A Vorto primitive for map structure

### LINKED ISSUE #222: A Vorto primitive for map structure
Map structures (sets of key/value pairs) are very common both in programming and in describing IoT devices. Vorto already has a sort of primitive for the other most commonly used data structire (list/array) with `multiple`. It would be convenient if Vorto provides a primitive for map structures as well.

erlemantos:


bundles/org.eclipse.vorto.editor.datatype/src/org/eclipse/vorto/editor/datatype/Datatype.xtext
```diff
@@ -111,6 +111,14 @@ ObjectPropertyType:
 	type = [Type|QualifiedName]
 ;
 
+ComplexPrimitivePropertyType:
+	DictionaryPropertyType
+;
+
+DictionaryPropertyType:
+	{DictionaryPropertyType} 'Dictionary' ('[' keyType = PrimitiveType ',' valueType = PrimitiveType ']')?
```
aedelmann: Why is the keytype and valuetype only restricted to PrimitiveTypes? 
I can imagine that a user would also like to define a constraint for Object Types , like in a ordinary map.What do you think?

PR #308, opened 2016-06-27
remove eclipse UI dependency in generator

hrohmer:
To use the generators within a headless OSGI container, we should avoid dependencies to UI bundles e.g. org.eclipse.ui


aedelmann:
Thanks! Have you checked if this generator is the only one with a UI dependency?

hrohmer:
Yes, all generators may run within our OSGI container now ...


Von: Alexander Edelmann <notifications@github.com>
Antworten an: eclipse/vorto <reply@reply.github.com>
Datum: Montag, 27. Juni 2016 19:05
An: eclipse/vorto <vorto@noreply.github.com>
Cc: Heinz Rohmer <heinz.rohmer@generative-software.com>, Author <author@noreply.github.com>
Betreff: Re: [eclipse/vorto] remove eclipse UI dependency in generator (#308)

Thanks! Have you checked if this generator is the only one with a UI dependency?

PR #336, opened 2016-07-12
Return error information if code generation fails, fixes #331

### LINKED ISSUE #331: Return error information if code generation fails
If errors happen when user generate code, 
Expect result: get some error information
Actual result: nothing responsed

gavinying:
Signed-off-by: Ying Shaodong <helloysd@gmail.com>


aedelmann:
Hi Shaodong, 
I saw that you are doing the exception handling in specific generator tasks. That would mean, if another generator not using the tasks that handle the error, is used, would not create this error log file , correct?

I would suggest to handle an unchecked exception in the generator client code which invokes the IVortoCodeGenerator and creates the log file and returns it. That way any code generator producing an error regardless of which generator tasks it uses, handles errors automatically. What do you think?

aedelmann:
@gavinying 
I have not received any pull request with that change, we have discussed in the planning. Are you facing any issue or is it unclear?

gavinying:
@aedelmann 
A little busy with other project today. Will fix it right now.

aedelmann:
Other project ?!? Let's discuss that on monday.

gavinying:
Created new pull request, see https://github.com/eclipse/vorto/pull/338

PR #338, opened 2016-07-15
Return error log when code generation fails; fixes #331;

### LINKED ISSUE #331: Return error information if code generation fails

gavinying:
Signed-off-by: Ying Shaodong <helloysd@gmail.com>

### DIFF
--- a/bundles/org.eclipse.vorto.codegen.ui/src/org/eclipse/vorto/codegen/ui/handler/CodeGeneratorTaskExecutor.java
+++ b/bundles/org.eclipse.vorto.codegen.ui/src/org/eclipse/vorto/codegen/ui/handler/CodeGeneratorTaskExecutor.java
@@ -54,10 +54,8 @@ public class CodeGeneratorTaskExecutor {
 				try {
 					generatedResult = ((IVortoCodeGenerator) codeGenerator).generate(informationModel,invocationContext);
 				} catch (Exception e) {
-					GenerationResultZip output = new GenerationResultZip(informationModel,codeGenerator.getServiceKey());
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.vorto.codegen.ui/src/org/eclipse/vorto/codegen/ui/handler/CodeGeneratorTaskExecutor.java
```diff
@@ -48,17 +50,19 @@ protected IStatus run(IProgressMonitor monitor) {
...
-					monitor.worked(1);
-					if (monitor.isCanceled()) {
-						monitor.done();
-						return Status.CANCEL_STATUS;
-					}
-
+					generatedResult = ((IVortoCodeGenerator) codeGenerator).generate(informationModel,invocationContext);
 				} catch (Exception e) {
-					e.printStackTrace();
+					GenerationResultZip output = new GenerationResultZip(informationModel,codeGenerator.getServiceKey());
```
aedelmann: The Eclipse Client Toolset should not have an error.log. Instead it would be ideal, if the message is outputted to the Eclipse Console using MessageDisplay.error()

gavinying:
I thought you want to make the generation process always be successful, so that the toolset and webservice would function the same way to users. 
Just double check, so you want toolset to cancel the generation and print error message if error happens, right?

aedelmann:
The Web service returns an error.log file back to the user while the toolset prints out the error in the console and aborts the generation.

aedelmann:
Looks great. Thanks!

PR #467, opened 2016-12-09
Fixes #458 - Add more metadata to the description of generated json f…

### LINKED ISSUE #458: Add more metadata to the description of generated json files
Reference function block ID as metadata description in Bosch IoT Things Generator

erlemantos:
…iles

Signed-off-by: Erle Czar Mantos <ErleCzar.Mantos@bosch-si.com>


aedelmann:
Hi @erlemantos 
You can actually retrieve the function block model much easier from the specific object passed to the template , e.g. configuration. you just need to call configuration.eContainer.eContainer which will give you the function block model and the model id. 
That way you do not change the context object inside the generator coz it is supposed to stay immutable.
Cheers, Alex

Issue #459, opened 2016-12-02
Bosch IoT things generator outputs folder structure matching things paradigm


aedelmann:

stefferber:
;-)

erlemantos:
Can you give a more detailed explanation for this issue? What things paradigm are we to follow?

erlemantos:
Given the XDK information model, would you like this kind of structure?

```
|-- feature
|   |-- accelerometer
|   |   `-- status
|   |       `-- status.schema.json
|   |-- alertnotification
|   |   |-- operations
|   |   |   |-- resetAlert-result.schema.json
|   |   |   `-- resetAlert.schema.json
|   |   `-- status
|   |       `-- status.schema.json
|   |-- barometer
|   |   |-- operations
|   |   |   |-- resetMinMaxValues-result.schema.json
|   |   |   `-- resetMinMaxValues.schema.json
|   |   `-- status
|   |       `-- status.schema.json
|   |-- gyroscope
|   |   `-- status
|   |       `-- status.schema.json
|   |-- humiditysensor
|   |   |-- operations
|   |   |   |-- resetMinMaxValues-result.schema.json
|   |   |   `-- resetMinMaxValues.schema.json
|   |   `-- status
|   |       `-- status.schema.json
|   |-- lightcontrol
|   |   |-- configuration
|   |   |   `-- configuration.schema.json
|   |   `-- status
|   |       `-- status.schema.json
|   |-- lightsensor
|   |   |-- operations
|   |   |   |-- resetMinMaxValues-result.schema.json
|   |   |   `-- resetMinMaxValues.schema.json
|   |   `-- status
|   |       `-- status.schema.json
|   |-- magnetometer
|   |   `-- status
|   |       `-- status.schema.json
|   |-- temperaturesensor
|   |   |-- operations
|   |   |   |-- resetMinMaxValues-result.schema.json
|   |   |   `-- resetMinMaxValues.schema.json
|   |   `-- status
|   |       `-- status.schema.json
|   |-- xdksensordevice
|   |   |-- configuration
|   |   |   `-- configuration.schema.json
|   |   `-- status
|   |       `-- status.schema.json
|   `-- xdksensordevice1
|       |-- configuration
|       |   `-- configuration.schema.json
|       `-- status
|           `-- status.schema.json
```

erlemantos:
If the above structure is satisfactory, the branch for modified code is in -> https://github.com/bsinno/vorto/tree/feature/vorto-459

[...]

PR #479, opened 2017-01-10
Feature/vorto 459

erlemantos:
Modified the generator to follow the convention set in this issue -- https://github.com/eclipse/vorto/issues/459


aedelmann:
Awesome. Looking great! 👍

PR #503, opened 2017-02-02
improvements to java client

erlemantos:
Improvements:

0. Renamed typo from oeg to org
1. Do the encoding of the query parameter in the client, instead of ModelQueryBuilder
2. Use Function.identity() instead of reinventing the wheel
3. Use functional notation on some functions
4. Add Eclipse EPL headers
5. Add maven profile for gpg signing, include source, and release to OSS Sonatype.


server/repo/repository-java-client/src/test/java/org/eclipse/vorto/repository/client/TestRepositoryClient.java
```diff
@@ -32,15 +47,35 @@
...
-import oeg.eclipse.vorto.repository.client.RepositoryClientBuilder;
-
 public class TestRepositoryClient {
 
 	private static final int TEST_PORT = 6666;
 	
 	@Rule
 	public WireMockRule rule = new WireMockRule(options().port(TEST_PORT));
 	
+	/*
```
aedelmann: Can this be removed ?

PR #509, opened 2017-02-06
Add plugin repo for eclipse CBI releases

erlemantos:
Signed-off-by: Erle Czar Mantos <ErleCzar.Mantos@bosch-si.com>


erlemantos:
This is to fix the current build error in Hudson

PR #556, opened 2017-05-16
Fixes #537 - Option to run backend generators as single bundle to save resources on target environment

### LINKED ISSUE #537: Option to run backend generators as single bundle to save resources on target environment
Micro service that can be configured to bundle up multiple generators and delegates incoming generation requests to in-process generators. 
This saves machine resources in target environments with limited memory and thus is not able to run generators individually.

Confirmations:
- Compound Generator Gateway service routing generation request to configured embedded generators
- Each configured generator is registered with the Vorto Repository

erlemantos:


server/generators/gateway/src/main/java/org/eclipse/vorto/codegen/gateway/config/SecurityConfig.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Component
+public class SecurityConfig {
+	
+	private static final Logger LOGGER = LoggerFactory.getLogger(SecurityConfig.class);
+
+	@PostConstruct
+	public void ignoreSSLCertificateVerification() {
+		SSLContext ctx = null;
+        TrustManager[] trustAllCerts = new X509TrustManager[]{new X509TrustManager() {
```
aedelmann: Not very secure , is it ?

PR #569, opened 2017-05-25
GitHub integration

erlemantos:
- Add support for Github oAuth provider
- Allow enabling/disabling of oAuth providers
- Github-authenticated users should go through registration screen


aedelmann:
Hi Erle, 
can you please have a look at the codacy result ? 
Thx.

PR #632, opened 2017-06-22
Fixes #625

### LINKED ISSUE #625: Models are displayed in Content assist that have been imported by other projects
Steps to reproduce:
1) open vorto.eclipse.org/editor
2) Create project
3) Create functionblock model
4) Import datatype from model browser, e.g. mytype
5) Create status property in fb model and open content assist for type

Expected: Only primitive types and project import type 'mytype' is displayed and can be used.
Actual: Other types are shown from other projects or even other users

shiv12095:


shiv12095:
@aedelmann 
This commit will break the original functionality where in all references were imported. 
Users will have to re import all their models from the repository again after this change.

PR #643, opened 2017-06-24
Displaying loader on front end while the file explorer is being rendered

shiv12095:

### DIFF
--- a/server/devtool/webeditor/src/main/resources/application-local.yml
+++ b/server/devtool/webeditor/src/main/resources/application-local.yml
@@ -3,7 +3,7 @@ vorto:
   repository:
     base:
       path: http://vorto.eclipse.org
-      
+
 http:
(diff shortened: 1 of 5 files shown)


server/devtool/webeditor/src/main/resources/application-local.yml
```diff
@@ -3,23 +3,23 @@ vorto:
...
 server:
   port: 9080
   contextPath: /editor
-  
+
 github:
   client:
-    clientId: ${github_clientid}
-    clientSecret: ${github_secret}
+    clientId: b75406b8428b88732920
```
aedelmann: You better not check in your credentials :)

shiv12095: Thanks for pointing that out. 
Removed these credentials and created a new commit

PR #668, opened 2017-06-29
Updated delete resource API request type from POST to DELETE

shiv12095:


aedelmann:
Hi Shiv. Do you check the permissions that you are only allowed to delete your own projects not anyone else's ?

PR #712, opened 2017-08-07
Vorto setup for contributors. Fixes DX-639.

deepagithub:
Signed-off-by: deepagithub <deepalakshmi.srinivasan@bosch-si.com>


aedelmann:
Great :)

PR #720, opened 2017-08-22
Initial commit for a simple Arduino Code Generator

omeili:


generators/org.eclipse.vorto.codegen.arduino/pom.xml
```diff
@@ -0,0 +1,224 @@
+<project>
```
aedelmann: Can you make your generator inherit from the parent, just as the other generators ?

PR #732, opened 2017-08-27
Initial commit for a Python client which uses MQTT to send data to an

omeili:
MQTT broker


aedelmann:
Great stuff. U might wanna add your generator to the generators feature.xml so it will be installed to the toolset via the eclipse update site.

aedelmann:
Another thought: would it make sense to add your generator to the existing mqtt generator as a generator task? So the generator would support java and python. A user would be able to choose the target language during generator config time.

omeili:
I have been thinking about that as well, I assume a user would only want to use one of them at a time, that selection however could also be a runtime configuration for the MQTT generator.
I also believe the two generators serve vastly different purposes: the MQTT seems to generate a backend component, while the Python generator generates device code

omeili:
Maybe it would make sense to characterize generators according to their purpose: documentation, device code, backend code, ...

PR #749, opened 2017-09-10
Changed unwrap to accept non-list arguments

omeili:


omeili:
Signed-off-by: <omi@ieee.rog>

omeili:
Signed-off-by: <omi@ieee.org>

PR #753, opened 2017-09-12
Fixed issue when multiple custom functions are defined in different

omeili:
function blocks that only the first functions was incorrectly
referenced.


aedelmann:
Hi Oli. Can u check your formatting config. From this pull request I cannot see your changes as it marks everything as New.

omeili:
I'd assume setting the following would do the trick on Windows, correct?
git config --global core.autocrlf true

aedelmann:
IMO, that should do the trick :)

PR #756, opened 2017-09-16
Added an object model for BLE GATT devices

omeili:


aedelmann:
Hi Oli. Wow nice. 
Can u make codacy happy as well add EPL header to the java sources. 

Thanks.

PR #764, opened 2017-09-30
- Allow for partial mapping of function blocks, targetplatform is null

omeili:
for unmapped functionblocks
- Corresponding changes in BLE functionality by checking for
targetplatform == "blegatt"

Apologies for the diff formatting of the crucial pieces


omeili:
Change is fairly straightforward: if there is no functionblock from the mapping, just retrieve the "naked" one from the repo, without mapping information. Users need to check for the targetplatform or just recognize that the corresponding properties/attributes aren't there.

PR #775, opened 2017-10-15
Generator for Alpwise BLE Stack and Generic BLE Utilities

omeili:
plus a small bonus fix for the Arduino Generator...


omeili:
Realized have of the commit is missing, resubmitting

PR #802, opened 2017-11-20
The property to change max file size for an upload has changed in Spring Boot

Ebolon:
See: [Stackoverflow](https://stackoverflow.com/questions/36386462/multipart-file-exceed-limit-exception/36386563#36386563) and [Spring Book Guide - Uploading Files](https://spring.io/guides/gs/uploading-files/)


aedelmann:
Thanks @Ebolon :)

PR #806, opened 2017-11-22
Fixes #DX-918 CLM configurations in pom.xml for webeditor, gateway an…

vlvanchin:
…d repo

Signed-off-by: val8sgp <vanchinathan.lokanathan@bosch-si.com>


aedelmann:
@vlvanchin 
there should not be any Bosch Tina related maven configurations in vorto.

PR #808, opened 2017-11-23
Using EcoreGenerator to generate Vorto Core API source code

Ebolon:
Changes:
- Removed all generated source code from org.eclipse.vorto.core
- Using src-gen folder
- Created workflow to generate code form ECore (GenerateModels.mwe2)
- Modified main pom.xml to include /target in mvn clean
- Modified manifest + pom to match dependency requirements for eclipse

Benefit: Update, extend Vorto DSL without massive code changes in repository.


aedelmann:
Hi @Ebolon 
very nice contribution :)

PR #871, opened 2018-03-29
Fix JSON Generator dependencies

Ebolon:
Yep, I already thought that `maven-bundle-plugin` could cause some trouble. Here I added all dependencies manually and removed the gen manifest stuff.


Ebolon:
Done.

PR #873, opened 2018-03-29
Updated dependencies and added default values to local conf

Ebolon:
The ModelHandle definition is missing in http://vorto.eclipse.org/v2/api-docs. Because of a bug in `springfox-swagger2`(see: https://github.com/springfox/springfox/issues/1282)

Currently the api doc is invalid. Error in Swagger:
screenshot_328

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -230,18 +230,18 @@
 			<dependency>
 				<groupId>io.swagger</groupId>
 				<artifactId>swagger-annotations</artifactId>
-				<version>1.5.6</version>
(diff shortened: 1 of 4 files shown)


aedelmann:
Hi @Ebolon 
can you test it with version 1.5.17 of the io.swagger and 2.6.1 of io.springfox and see if it works with these versions too ? 
Found existing CQs in Eclipse for those. Thanks.

Ebolon:
Yep, solves also the problem with those versions. Done.

PR #876, opened 2018-04-12
Fix source format in eclipse for functionblock, etc.

Ebolon:
For me the `Source -> Format` in Eclipse had no effect for a functionblock. After some research I have found in the Xtext unit tests (see [Link](https://github.com/TemenosDS/xtext-core/blob/3ba9c9fa57932cb81cac113e0c6a0d4ff9daf553/org.eclipse.xtext.tests/src/org/eclipse/xtext/parsetree/formatter/FormatterTestLanguageRuntimeModule.java)) that `bindIFormatter` needs a return statement.


aedelmann:
:)

PR #920, opened 2018-05-07
Feature/gdpr toolset share improvement

erlemantos:
- Add detailed instruction on dialog when you click Share on the toolset
- Changed to "Created by You" and "Created by other User" instead of showing hashed usernames.

### DIFF
--- a/server/repo/repository-web/src/main/resources/static/index.html
+++ b/server/repo/repository-web/src/main/resources/static/index.html
@@ -1,106 +1,128 @@
 <!DOCTYPE html>
 <html lang="en" ng-app="repository">
-  <head>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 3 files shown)


server/repo/repository-web/src/main/resources/static/index.html
```diff
@@ -1,106 +1,128 @@
...
+					<li ng-show="authenticated && userInfo.isRegistered === 'true'"><a
+						href="./#/upload">Share</a></li>
+					<li ng-show="authority == 'ROLE_ADMIN'"><a href="./#/manage">Manage</a></li>
+					<li><a href="./#/generators">Generators</a></li>
+					<li><a href="http://www.eclipse.org/vorto">About Vorto</a></li>
+					<li><a href="./#/api">REST API</a></li>
+				</ul>
+				<div id="navbar-right" class="navbar-right">
+					<ul class="nav navbar-nav">
+						<li ng-show="authenticated"><a id="userWelcome"
```
aedelmann: I think it makes sense here to also check if the person is registered , otherwise it shows the user despite the fact that he has not yet agreed to T&C.

Issue #823, opened 2017-12-14
Overriding properties inline


Ebolon:
Like discussed in #822 it should be possible to override some attributes inline in a functionblock/entity.

Here some examples:
_Example1:_
```
namespace com.ipso.smartobjects

functionblock Distance {

   status {
       mandatory sensor_value as float
       optional sensor_units as string 
       ...    
   }
}
```
Information model using the IPSO model and overriding it:
```
using com.ipso.smartobjects.Distance;0.0.1

infomodel MyDistance100MeterDevice {

   functionblocks {
       distance as Distance with {
            mandatory sensor_value as float <max:100>
       }
   }
}
```

_Example2:_
```
entity EntityA {
	mandatory valueA as int
	optional valueB as int
}
```
```
using de.vorto.EntityA;0.0.1

functionblock FunctionblockA {
   status {
       mandatory entityA as EntityA
       optional unit as string 
   }
}
```
```
using de.vorto.FunctionblockA;0.0.1

infomodel DeviceA {
   functionblocks {
       distance as FunctionblockA with {
            mandatory entityA as EntityA with {
            	valueA as int <MAX 100>
            }
       }
   }
}
```

aedelmann:
I like that change. I think that would make the DSL very powerful because information models would have a way to specialize abstract and re-usable functionblocks in an inline way. Would you dare to extend the Informatioin Model Definition DSL to make this change  :) ?

Ebolon:
> Do I understand correctly that by this I can no longer rely on that a functionblock always contains a well defined (and versioned) contract but that it depends on the context?

It depends on your information model (IM) which is the base for all Vorto models. The Vorto framework currently only uses the generated Ecore classes as they are. This should change in the future to enable the inheritance feature of a functionblock (see #796). As a user of the Vorto framework you need not to care about any extensions of a functionblock. In the future there will be a extended Ecore model which can be used by the generators.

> What is the point of a functionblock then if I cannot rely that it is the same when used in different infomodels?

The benefit is that you can change or extend specific parts like the anonymous class concept known from programming languages. The point of a functionblock stays the same you have a common base for development.

Maybe we can align how you use the Vorto DSL. Have you a scenario in mind where this language feature could be a problem?

thjaeckle:
Yes, I have scenarios in mind where this is a problem :)

Coming from the "digital twin" (Eclipse Ditto) side where the information models and functionblocks come to live with concrete instances assume you have:

[...]

PR #932, opened 2018-05-08
Added extend feature to information model dsl

Ebolon:
As discussed in #823 it can be helpful to extend/override attributes of a function block in an information model. The `with` keyword is already used in another context to give a property additional attributes. Thats why the keyword `extend` is used. In and `extend` block all features of a function block can be used. Other like the proposed idea the fields status, etc. are retained because property names between sections are not unique. 

Example:
```
infomodel SenseHat {
    functionblocks {
        gyrometer as Gyrometer extend { 
            status {
                offline as boolean
            }
            operations {
                set_active()
            }
        }
    }
}
```

The feature could be implemented similar for data types in a function block in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ generated-sources
 *.orig
 
 devtool-project-repository/
+
+*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,3 @@ generated-sources
 *.orig
 
 devtool-project-repository/
-
-*.iml
--- a/bundles/org.eclipse.vorto.editor.functionblock.tests/src/org/eclipse/vorto/editor/functionblock/tests/validator/FBExtendsTest.xtend
+++ b/bundles/org.eclipse.vorto.editor.functionblock.tests/src/org/eclipse/vorto/editor/functionblock/tests/validator/FBExtendsTest.xtend
@@ -1,0 +1,56 @@
+package org.eclipse.vorto.editor.functionblock.tests.validator
+
+import org.eclipse.xtext.junit4.AbstractXtextTests
+import org.eclipse.xtext.junit4.validation.ValidatorTester
+import org.eclipse.vorto.editor.functionblock.validation.FunctionblockValidator
+import org.eclipse.vorto.editor.functionblock.FunctionblockStandaloneSetup
+import org.eclipse.vorto.core.api.model.functionblock.FunctionblockFactory
+import org.eclipse.vorto.core.api.model.datatype.DatatypeFactory
+import org.eclipse.vorto.core.api.model.datatype.PrimitiveType
+import org.eclipse.vorto.editor.functionblock.validation.SystemMessage
+import org.junit.Test
+
+class FBExtendsTest extends AbstractXtextTests {
+		
+	private ValidatorTester<FunctionblockValidator> tester;
+	
+	def override void setUp() throws Exception {
+		super.setUp();
+		with(FunctionblockStandaloneSetup);
+		var validator = get(FunctionblockValidator);
+		tester = new ValidatorTester<FunctionblockValidator>(validator, getInjector());
+	}
+	
+	@Test
+	def test_IM_extend_datatype_conflict() {
+		var fbModel = FunctionblockFactory.eINSTANCE.createFunctionblockModel()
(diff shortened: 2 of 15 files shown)


Ebolon:
This is a tough decision. I see to potential options:

Option 1:
It could be nice that it is not required to write all "keywords" again. A with-block could be overwritten with `with {}` and a comment `""`. But if a with-block is overwritten all previous defined variables should be deleted. But then we should behave the same way when `extends` is used.

Option 2:
The property needs to be defined again completely.

For me option 2 would fits best. We assume that most people have used a OOP language. This would give the same experience. Maybe a bit more verbose but easier to understand.

_Example for option 2:_
```
namespace com.ipso.smartobjects

functionblock Distance {

   status {
       mandatory sensor_value as float with {readable : true} "Last or Current Measured Value from the Sensor"
       ...    
   }
}
```
Adding constraints to `sensor_value`:
```
using com.ipso.smartobjects.Distance;0.0.1

infomodel MyDistance100MeterDevice {

   functionblocks {
       distance as Distance extend {
            status {
                mandatory sensor_value as float with {readable : true} <MIN:0, MAX:100> "Last or Current Measured Value from the Sensor"
            }
       }
   }
}
```

aedelmann:
Hi @Ebolon 
Looking at these two options, I think so too, that option 2 is better. But the validation has to make sure that you are not able to override the datatype, as OOP languages would not allow that either. I can see, that your Pull Request is not handling this yet. 
Another thing in the context of this: how are these overridden  function block properties processed in generators etc ? My idea would be to wrap the current EMF Model into a more convenient domain model that is easier to be processed by other components, like generators or the repository. It could deal with function block inheritance etc and give you a consolidated view on the properties. what do you think ? 

Btw: travis ci is failing because of failing unit tests in repository-java client. Can you take a look :-)  ?

aedelmann:
Hi @Ebolon 
Would you mind adding the necessary DSL validator that validates if the extended properties do not override/clash with the definition from the base function block ?

Ebolon:
Yep, will have a look next week.

aedelmann:
Cool. Thanks @Ebolon 
I will try it out this week and let you know.

aedelmann:
Looking at your validation, it seems that it is not checking if there is a clash of a constraint and with statement ? It could be that the base property defines eg a MAX constraint which the overridden property should not be able to override. Same applies for with statement

Ebolon:
For the constraints I have the idea that it would be feasible to tight boundaries. Here an example:
```
functionblock BlockA {
    configuration {
        mandatory value as int with {readable: true} <MIN 1, MAX 500>
    }
}

functionblock BlockB extends BlockA {
    configuration {
        mandatory value as int with {readable: true} <MIN 10, MAX 100>
    }
}
```
How do you think about that?

I am not 100% sure how the validation should behave for the with-block. For example:
```
functionblock BlockA {
    configuration {
        mandatory value as int with {readable: true}
    }
}

functionblock BlockB extends BlockA {
    configuration {
        mandatory value as int with {writable: true}
    }
}
```
Should it be ok to change `readable`, `writable`,  `eventable`?

When there is a clash between constraints an with-block?

aedelmann:
@Ebolon 
That would work for min and max. So for the others (regex, ..) the validation would just fail ? 
Regarding with block: Any existance of _with_ definitions would not be allowed to be overridden by base definitions. That includes readable, writable, eventable and measurementUnit.

Ebolon:
Here is my proposal:

|constraint | overridable | comment |
|-----------|------------|-----------|
| MIN | YES | parent value <= value |
| MAX | YES | parent value >= value |
| STRLEN | YES | parent value >= value |
| REGEX | NO | |
| MIMETYPE | NO | |
|SCALING | NO | |
| DEFAULT | NO | |
| NULLABLE | YES | if parent is true it can be overwritten with false |

For the with-block I agree. If any keyword is already defined it can not be overwritten.

Ebolon:
@aedelmann: I have added the intended validation functions for property attributes and constraints. Maybe you can have a look if they work like you would expect. I create the necessary units test afterwards.

Ebolon:
@aedelmann as discussed you can close this pull request. I will create a new one and remove the inline extend and including only the `extends` validation.

PR #949, opened 2018-05-31
Added validation for FB extends

Ebolon:
I have implemented that if a value is overwritten not all property attributes and constraints have to be written again. They are inherited from the parent FB. I am not sure if this would create more confusion and complications about the behaviour. The editor shows an error message if this rule is violated. Unit test currently not complete.

I am open for a discussion. How do you feel about this @aedelmann? Yes/No would be fine. The other option would be to require all attributes and constraints to be written again. I am a bit undecided here.

**Here some examples:**
```
mandatory xValue as float with { readable: true} 
```
Can be overwritten with:
```
mandatory xValue as float <MAX 100>
```
Evaluated as:
```
mandatory xValue as float with { readable: true} <MAX 100>
```
**Second example:**
```
mandatory xValue as float <MIN 10, MAX 100>
```
Can be overwritten with:
```
mandatory xValue as float <MAX 99>
```
Evaluated as:
```
mandatory xValue as float <MIN 10, MAX 99>
```
**Third example:**
```
mandatory xValue as float with { readable: true} 
```
Can be overwritten with:
```
mandatory xValue as float with { writable: true} 
```
Evaluated as:
```
mandatory xValue as float with { readable: true, writable: true}
```

### DIFF
--- a/bundles/org.eclipse.vorto.editor.datatype/src/org/eclipse/vorto/editor/datatype/Datatype.xtext
+++ b/bundles/org.eclipse.vorto.editor.datatype/src/org/eclipse/vorto/editor/datatype/Datatype.xtext
@@ -63,7 +63,7 @@ EnumLiteral:
 ;
 
 Property:
-	(presence = Presence)? (multiplicity ?= 'multiple')? name = ID 'as' type = PropertyType
+	(extension ?= 'extension')? (presence = Presence)? (multiplicity ?= 'multiple')? name = ID 'as' type = PropertyType
 	('with' '{' propertyAttributes+=PropertyAttribute (',' propertyAttributes+=PropertyAttribute)* '}')?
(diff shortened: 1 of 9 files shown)


aedelmann:
Hi @Ebolon 
I think it's missing some keyword, like in java ' super' or in protobuf 'extendion'. 

How about this: 

Base FB:
`mandatory xValue as float with { readable: true} `

Extended FB:
`extension xValue with <MAX 100>`

This would remove the redundancy completely and clearly mark that this property is extended. 
What do you think?

Ebolon:
Sorry, should have mentioned that this is not finish. Working on further validation of the `extension` keyword. I will open a new pull request.

aedelmann:
Hi @Ebolon 
No problem. Btw: I added the new Vorto web site (Hugo - based), so you can add your tutorial also for 'Vorto Flow' :)

PR #959, opened 2018-06-13
Fixes #956 - Allow FB section names as property name

### LINKED ISSUE #956: Keywords as a part of the Vorto models
Initially, this topic has been discussed in this [issue](https://github.com/eclipse/vorto/issues/908#issuecomment-386564325).

In our model, there is a "status" property where the asynchronous operation progress is reported. It looks like:
```
...
status {
  mandatory status as FirmwareUpdateStatus
}
...
```
The issue is that "status" is a keyword and this model doesn't compile. As a workaround, we are using:
```
mandatory ^status as FirmwareUpdateStatus
```
but that's a good point for improvement.

Ebolon:
Now `status`, `events`, `operations`, `configuration` and `fault` are allowed as property name. 
Lets discuss which other keywords make sense to add in a new issue. If I have time I like to create a list of all available keywords.


aedelmann:
Cool! 
There is some IP validation problem when I wanna merge your PR. Can u check ?

Issue #965, opened 2018-06-15
Operation name cannot be same as custom type (Enum / Entity) name


e-grigorov:
I got this error message with this model:
```
using com.mycompany.type.Record ; 1.0.0

functionblock Recorder {
	operations {
		mandatory record() returns Record
	}
}
```
It's somehow strange limitation. What is the reason to have this restriction?

Ebolon:
I think you mean this error: "Operation name cannot be same as custom type (Enum / Entity) name". Yes this is strange. Would suggest to remove  the validation check `checkOpNameAgainstEntityName`. @aedelmann ?

PR #966, opened 2018-06-15
Fixes #964, #965 - Remove validation checks 

### LINKED ISSUE #964: Recursive definitions
I am trying to define a model similar to:
```
entity Exception {
	optional cause as Exception
}
```
Currently, it fails with: Object property type has circular reference

What is the best way to model this?

Ebolon:
Remove validation checks for datatype property recursion and same name validation for function block operation name.


aedelmann:
I’d be careful removing the recursion check as all generators need to be verified that they handle the recursion. Also the vorto perspective has to handle it. Can u make these changes ?

Ebolon:
Yep, I take care of this. Maybe it is now time for some unit tests for the generators :)

aedelmann:
HI @Ebolon 
Can you check your PR. It says Merge conflicts. I did quite a bit of changes lately, where I put the editors in a different folder structure. Also I removed the Eclipse Toolset, as we will be not supporting the Eclipse plugins any longer. Instead, I added an initial version of a simple web-based editing functionality in the Vorto Repository (using simple javascript ACE editor).

Ebolon:
Closing this for now. I open a new PR when I got the time to take care of this change.

PR #978, opened 2018-06-21
Feature/attachments

erlemantos:
Add API for adding attachments to a model

### DIFF
--- a/repository/repository-api/src/main/java/org/eclipse/vorto/repository/api/attachment/AttachResult.java
+++ b/repository/repository-api/src/main/java/org/eclipse/vorto/repository/api/attachment/AttachResult.java
@@ -4,25 +4,21 @@ import org.eclipse.vorto.repository.api.ModelId;
 
 public class AttachResult {
 	private boolean success;
-	private ModelId modelId;
-	private String filename;
-	private String urlLink;
+	private Attachment attachment;
 	private String errorMessage;
 
 	public static AttachResult success(ModelId modelId, String filename) {
-		return new AttachResult(true, modelId, filename, Attachment.getLink(modelId, filename), null);
+		return new AttachResult(true, Attachment.newInstance(modelId, filename), null);
 	}
 
 	public static AttachResult fail(ModelId modelId, String filename, String errorMessage) {
-		return new AttachResult(false, modelId, filename, null, errorMessage);
+		return new AttachResult(false, null, errorMessage);
 
 	}
 
-	public AttachResult(boolean success, ModelId modelId, String filename, String urlLink, String errorMessage) {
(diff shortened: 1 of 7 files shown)


repository/repository-server/src/main/java/org/eclipse/vorto/repository/core/IModelRepository.java
```diff
@@ -127,4 +127,32 @@
...
+	/**
+	 * Attaches the given file to the model
+	 * 
+	 * @param modelid The modelId where to attach the file
+	 * @param fileName the filename
+	 * @param content the content of the file
+	 * @param userContext the user context
+	 * @return
+	 */
+	boolean attachFile(ModelId modelid, String fileName, byte[] content, IUserContext userContext);
```
aedelmann: You can use FileContent class to pass to this function instead of filename and content individually. The other methods behave the same.

repository/repository-server/src/main/java/org/eclipse/vorto/repository/core/IModelRepository.java
```diff
@@ -127,4 +127,32 @@
...
+	
+	/**
+	 * Gets the list of attachments for the model
+	 * 
+	 * @param modelId
+	 * @return
+	 */
+	List<String> getAttachments(ModelId modelId);
```
aedelmann: What does this method return ? Attachment file names , keys, ? Please rename the method better.

repository/repository-server/src/main/java/org/eclipse/vorto/repository/core/IModelRepository.java
```diff
@@ -127,4 +127,32 @@
...
+	
+	/**
+	 * Gets the content of the attachment
+	 * 
+	 * @param modelid The model id where the file was attached
+	 * @param fileName the filename of the attachment
+	 * @return
+	 */
+	Optional<byte[]> getAttachmentContent(ModelId modelid, String fileName);
```
aedelmann: Please return FileContent class

repository/repository-api/src/main/java/org/eclipse/vorto/repository/api/attachment/AttachResult.java
```diff
@@ -0,0 +1,69 @@
+package org.eclipse.vorto.repository.api.attachment;
+
+import org.eclipse.vorto.repository.api.ModelId;
+
+public class AttachResult {
+	private boolean success;
```
aedelmann: AttachmentResult contains fields that are already in the Attachment object. I would suggest to have a reference to attachment objects.

aedelmann:
Hi @erlemantos  
I added some improvements for your code.

PR #984, opened 2018-06-27
Add a way for the generator service to be able to authenticate agains…

erlemantos:
…t the server.


aedelmann:
@erlemantos 
As discussed, IMO the token should be passed to the generator service in the http header instead of in the URL.

aedelmann:
@erlemantos There are some conflicts ? Can you check that as well as the Codacy issues ?

erlemantos:
I'm checking it already

PR #1020, opened 2018-08-06
Feature/migration

erlemantos:
This is in preparation for user migration features which we might need in the future.


aedelmann:
@erlemantos 
Does the Update Account Modal pop up immedialtely after he logged in for the first time without any user intervention or how is the behaviour?

erlemantos:
The Account Modal is similar to the Terms and Condition in behaviour. It popups after login once the following condition occurs:  the lastUpdated of the user is BEFORE server.config.updateDate and the current date is AFTER the server.config.updateDate config

PR #1040, opened 2018-08-16
Fixes #1037 - Added mandatory and optional to keyword list

### LINKED ISSUE #1037: mandatory/optional cannot be used as property names
Recently, we switched to M4 build and got some new fails in our TCs. The problematic scenario is a property with a name: "mandatory" or "optional" i.e.
```
functionblock Testable {
	status {
		mandatory optional as boolean with {
			eventable : true
		} <NULLABLE false>
	}
}
```
or
```
functionblock Testable {
	status {
		mandatory mandatory as boolean with {
			eventable : true
		} <NULLABLE false>
	}
}
```
org.eclipse.vorto.core.api.model.datatype.Property.getName() returns null but should return "optional" or "mandatory" as a name.

Ebolon:


aedelmann:
Yay. You are back to contribute :) 
Btw: do u still have that contribution for defining extensions in info models ? If yes , can u merge it ?

Ebolon:
:-) 

You would like to open #932 again? Is this ok for the ditto team?

aedelmann:
@Ebolon 
I created a new issue  #1041 :)

PR #1105, opened 2018-10-15
Feature/multitenancy alpha

erlemantos:
First cut for adding multitenancy support to Vorto

### DIFF
--- a/repository/repository-web/src/main/resources/static/js/app.js
+++ b/repository/repository-web/src/main/resources/static/js/app.js
@@ -57,7 +57,7 @@ repository.config([ "$routeProvider", "$httpProvider", function($routeProvider, 
         githubEnabled: false,
         eidpEnabled: false,
         authenticatedSearchMode: false,
-        singleTenantMode: true
+        tenant: "default"
     };
     
     $rootScope.modelId = function(namespace,name,version) {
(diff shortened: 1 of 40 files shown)


repository/repository-web/src/main/resources/static/js/app.js
```diff
@@ -51,10 +51,13 @@ repository.config([ "$routeProvider", "$httpProvider", function($routeProvider,
 
     $rootScope.user = null;
     
+    $rootScope.tenant = "default";
```
aedelmann: Hi @erlemantos 
What would be the implication, if the tenant is another , other than "default" ? Would this UI still work ?

erlemantos: That's just initialization code. It will also be the value of the tenant if Vorto is configured as singleTenantMode.

I imagine the value of that variable will be replaced once a user logs in, in multitenant mode. However, we still don't have integration with any OAuth provider that is multi-tenant capable. So, I still haven't included that part of tenant-switching.

aedelmann: Understand. What would we need to do to make the UI tenant agnostic, meaning that the tenant gets injected into UI by the backend? I understand we do not have the integration with Oauth provider yet, to actually read the tenant information but maybe there is another way, like single tenant strategy ?

PR #1129, opened 2018-11-02
Feature/server xtext validation

erlemantos:


erlemantos:
1. There are already several test cases that are directly invoking it. I'm assuming that if all those test cases passed, the model parser is working as before. Yes, we can add a test just for the model parser but I don't really see the point since almost all test invokes it.
2. There are no API breaks, just additional functionalities. The IPSO importer was creating invalid models which was caught by the xtext validator, but I already fixed that.

erlemantos:
Added test for ModelParser. The last test shows a validation that only xtext can give.

PR #1133, opened 2018-11-07
Fix DX-1896 since ModelId was moved to new package

vlvanchin:


aedelmann:
Pull Requests content misleading from title.

PR #1165, opened 2018-11-16
Feature/diagnostics

erlemantos:


repository/repository-core/src/main/java/org/eclipse/vorto/repository/web/diagnostics/DiagnosticsController.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping(value = "/rest/{tenant}/diagnostics")
+public class DiagnosticsController {
+
+	@Autowired
+	private JcrModelRepository modelRepo;
+	
+	@RequestMapping(method = RequestMethod.GET)
```
aedelmann: The diagnose endpoint is not secured by a specific role ?

erlemantos: Yeah, I forgot that one.

aedelmann:
Hi @erlemantos 
I created an issue for this diagnostics functionality: #1166 
Can you check if the description fits to what is implemented ?

erlemantos:
It should. Compliance to Vorto meta model is ensured by model validation checks. There's a check for references to, and references from. There's a metadata test if metadata conforms to model.

aedelmann:
Awesome. I add this additional info in the issue for tracking.

PR #1254, opened 2018-12-10
[1250] Fix error handling in Generators Health check

georgschmidtdumont:
By default the Error Handler for the Spring Rest-Template throws an
exception when the status code in a response is not 2xx.
Hence the check whether a Generator is available is changed from
checking the status code to handling the exception thrown by the
Rest-Templates error handler.

Signed-off-by: Schmidt-Dumont Georg (BCI/ESW24) <georg.schmidt-dumont@de.bosch.com>

### DIFF
--- a/repository/repository-core/pom.xml
+++ b/repository/repository-core/pom.xml
@@ -229,6 +229,13 @@
 			<artifactId>modeshape-unit-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		
+		<dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-starter-contract-stub-runner</artifactId>
+            <version>2.0.2.RELEASE</version>
+            <scope>test</scope>
+        </dependency>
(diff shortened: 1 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=1250

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=1250

aedelmann:
Hi @georgschmidtdumont 
I see that you do not have any integration tests written for the health. Would you add that ? 
I guess it could test the following:
1) Repository without any generators, to check if the repository is up
2) Repository with registered generator,  to check that repository is up but generator is not reachable

What do you think ?

georgschmidtdumont:
Hi @aedelmann,

that is what the tests in the GeneratorsHealthCheckTest class cover. I don't see any advantage in creating any additional tests.

PR #1414, opened 2019-02-22
Added tutorial about using Eclipse Vorto Plugins

aedelmann:


andrew-from-pretoria:
The Vorto Eclipse Toolset Plugins have been deprecated. 

"The Eclipse Vorto IoT Toolset Plugins are no longer supported. Instead you can simply use the Cloud-based Editors in the Vorto Repository." 

(Ref https://github.com/eclipse/vorto/blob/master/docs/release-notes.md )

PR #1423, opened 2019-03-08
Nodejs mapping library

someshku:
Vorto-mapping-engine in nodejs


someshku:
Will redo after checking licenses of used libraries.

PR #1536, opened 2019-05-07
Fixes issue-1534, added absolute image urls in readme

somesh-kumar:
Signed-off-by: Somesh Kumar <someshkumarsharma.sks@gmail.com>

### DIFF
--- a/mapping-engine-nodejs/README.md
+++ b/mapping-engine-nodejs/README.md
@@ -48,26 +48,26 @@ Mapping adds device specific information to an Information Model. Since the repr
 
 To create a mapping go to your newly created model and press the **Create Mapping Spec** Button.
 
-![create mapping spec button](./images/create_mapping_key.png)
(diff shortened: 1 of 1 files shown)


mapping-engine-nodejs/README.md
```diff
@@ -48,34 +48,34 @@ Mapping adds device specific information to an Information Model. Since the repr
 
 To create a mapping go to your newly created model and press the **Create Mapping Spec** Button.
 
-![create mapping spec button](./images/create_mapping_key.png)
+![create mapping spec button](https://github.com/eclipse/vorto/raw/master/mapping-engine-nodejs/images/create_mapping_key.png)
```
aedelmann: We should not put absolute paths here as it would not work on different branches. Please put relative paths. Take a look at other tutorial examples how to do that.

somesh-kumar: The readme.md was already using relative paths, unfortunately the relative paths do not resolve for the package homepage in npm(https://www.npmjs.com/package/vorto-mapping-utility), hence the change.

PR #1677, opened 2019-06-21
Feature/issue 1612 elastic search

erlemantos:
Prerequisites for this pull request to work-

(1) The account who owns the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY in TravisCI must be given permission to access the AWS ElasticSearch instance 
(2) A -Dserver.config.skipSslVerification=true must be added to the maven build command, for example
``
mvn -Dserver.config.skipSslVerification=true clean install
``


erlemantos:
Apparently, Travis CI doesn't allow environment variables defined in Repository Settings to be handed over to builds based on pull request from forks (like bsinno/vorto).

https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings

> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.

Travis CI guys also know about this [issue](https://github.com/travis-ci/travis-ci/issues/1301) and they don't have a good solution for it. 

If you think about it, it really is a security hole if they allow it. If I want to steal credentials, I can easily fork the project, and in some java unit test, email the credential to me, then PR it to the original project. Since the build runs automatically during PR, the test would get triggered and if I have access to the credentials in that unit test, I'd get the credentials in the email.

A way to trust "some" forks would be a workaround to this.

@aedelmann Can you try restarting the build from the Travis CI console? I want to see if it produces a different behaviour.

One possible short term solution would be to create a mock of the IndexingService that's backed by a local hash map instead of AWS elastic search, and use it during integration test, until we find a better solution. @aedelmann  Would you agree to this?

repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/indexing/IndexedModelInfo.java
```diff
@@ -0,0 +1,121 @@
...
+ * https://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.vorto.repository.core.indexing;
+
+import java.util.Map;
+import org.eclipse.vorto.model.ModelId;
+import org.eclipse.vorto.model.ModelType;
+import org.elasticsearch.search.SearchHit;
```
aedelmann: Hi @erlemantos 
Your model class contains a technical dependency to Elasticsearch. That way your interface IIndexService is bound to it as well. I would suggest to remove this and replace it a neutral representation. Your Elastic Service Implementation would then map the neutral and vice versa.

repository/repository-core/pom.xml
```diff
@@ -246,6 +251,20 @@
             <version>2.0.2.RELEASE</version>
             <scope>test</scope>
         </dependency>
+        
+        <!-- Elastic search indexing service (start) -->
+        <dependency>
```
aedelmann: Hi @erlemantos 
What is this dependency for ?

repository/repository-server-config/pom.xml
```diff
@@ -80,7 +87,21 @@
 			<artifactId>commons-compress</artifactId>
 			<version>1.18</version>
 		</dependency>
-
+		
+		<!-- AWS Elastic Search configuration (start) -->
+		<dependency>
+		    <groupId>com.amazonaws</groupId>
+		    <artifactId>aws-java-sdk-core</artifactId>
+		    <version>1.11.574</version>
```
aedelmann: Possible to use 1.11.392 instead ? as pre approved CQ available

repository/repository-server-config/pom.xml
```diff
@@ -80,7 +87,21 @@
...
+		</dependency>
+		
+		<dependency>
+		    <groupId>com.github.awslabs</groupId>
```
aedelmann: I cannot find any GitHub repository with the sources for the specified version. Can you please comment in the PR where you get it from ?

repository/repository-core/pom.xml
erlemantos: A dependency to the elastic core and dependency to the high level client of Elastic cluster

repository/repository-server-config/pom.xml
erlemantos: Can try.

repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/indexing/IndexedModelInfo.java
erlemantos: The SearchHit is there as part of a convenience function, not as part of the model. It can be moved in a utils somewhere. A factory is overkill, I think.

repository/repository-server-config/pom.xml
erlemantos: This is what I talked about in the email I sent last friday on vorto-dev. This artifact is not released in Maven central, hence the jitpack workaround.

It comes from this repo.

https://github.com/awslabs/aws-request-signing-apache-interceptor

Jitpack

https://jitpack.io/#awslabs/aws-request-signing-apache-interceptor

The version, I think, is lifted from the last github commit signature. Jitpack builds artifacts on demand.

The alternative to using this is to actually copy the file and put it in our project. I think we'll have an IP problem with that approach.

repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/indexing/IndexedModelInfo.java
aedelmann: Well designing it like you have currently, would make your API and model classes depend on elastic search. 
As a proposal:  I would create a separate maven module, lets say "repository-search-elastic" , which is the only module having a dependency to elastic search classes. 

The repository-core module, would merely define the API which is used by other core components and is completely independent of elastic search. 
At configuration time, you would bind the repository with the elastic search component together.

repository/repository-core/pom.xml
aedelmann: I see.

repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/indexing/IndexedModelInfo.java
erlemantos: We can definitely add that as an improvement.

erlemantos:
Currently, this is the one that's causing the error --

The command "remark -i .remark_ignore -f -u validate-links ." exited with 1.

aedelmann:
@erlemantos 
Can you resovle the conflict with the jenkins file , that was from your other PR ?

erlemantos:
> @erlemantos
> Can you resovle the conflict with the jenkins file , that was from your other PR ?

Will do

aedelmann:
Hi @erlemantos 
Have u included back the unit tests in the travis script ?

erlemantos:
> Hi @erlemantos
> Have u included back the unit tests in the travis script ?

Not yet. I will try to look for that particular incantation as my google-fu has been failing me for hours now.

aedelmann:
Hi @erlemantos 
As discussed, 
please provide a "SimpleIndexService" Implementation which searches all repositories without using elastic search. 
The INT Tests in the repository-server package should be configured using this simple search.

The AWS DEV and PROD environment shall use the elastic service implementation.

This has the benefits:
1. Set up of local repository can be easily done without the need to use elastic search
2. The integration tests would run with the simple search service and therefore the travis CI can run them

erlemantos:
> Hi @erlemantos
> As discussed,
> please provide a "SimpleIndexService" Implementation which searches all repositories without using elastic search.
> The INT Tests in the repository-server package should be configured using this simple search.
> 
> The AWS DEV and PROD environment shall use the elastic service implementation.
> 
> This has the benefits:
> 
> 1. Set up of local repository can be easily done without the need to use elastic search
> 2. The integration tests would run with the simple search service and therefore the travis CI can run them

After the model visibility pull request is merged, I will get on this, so I can merge and add this in one go.

aedelmann:
Merged publish story with elastic search. Closing this PR

PR #1819, opened 2019-07-19
1806 swagger api vorto website

MayilKannanKumarBosch:


aedelmann:
Is it possible to do that change only for website meaning where the website references the open api yml directly with this change instead of loading it from GitHub ? After your test u can revert it back.

MayilKannanKumarBosch:
> Is it possible to do that change only for website meaning where the website references the open api yml directly with this change instead of loading it from GitHub ? After your test u can revert it back.
> 
> I would then close this PR. And u make a PR on website GitHub project.

I think its better if we test this way...once it works in development we can revert it...pls let me know what u think...bcoz anyway we have a PR for the CORS changes...

aedelmann:
Hi @MayilKannanKumarBosch 
I tested the changes on dev with current web site. It works. No need for this PR. Hence closing it.

Issue #1802, opened 2019-07-15
Non org.eclipse.vorto function blocks show up as abstraction during model creation


aedelmann:
Steps to reproduce
1. Login with Bosch or GitHub id as user with model creator permission
2. Create new information model with 'Create Model'
3. During model creation, the Vorto abstraction wizard page shows. Select 'org.eclipse.vorto'
4. Open Drop down 
Expected: Drop down only shows function blocks in org.eclipse.vorto namespace
Actual: All public released function blocks are displayed

erlemantos:
I'll take this

PR #1821, opened 2019-07-22
Fix issue #1802 - Non org.eclipse.vorto function blocks show up as ab…

erlemantos:
…straction during model creation


aedelmann:
Hi Erle. Can u create PR again with latest change from dev ? Needed to fix sth in Travis script

PR #1883, opened 2019-08-16
Junit tests

MayilKannanKumarBosch:


generators/generator-eclipseditto/src/test/java/org/eclipse/vorto/codegen/ditto/java/EclipseDittoGeneratorTest.java
```diff
@@ -0,0 +1,96 @@
...
+				.addFile(getClass().getClassLoader().getResourceAsStream(
+						"dsls/ColorLight.type"), ModelType.Datatype)
+				.addFile(getClass().getClassLoader().getResourceAsStream(
+						"dsls/Light.type"), ModelType.Datatype)
+				.addFile(getClass().getClassLoader().getResourceAsStream(
+						"dsls/Brightness.type"), ModelType.Datatype)
+				.addFile(getClass().getClassLoader().getResourceAsStream(
+						"dsls/Coffee_Enum.type"), ModelType.Datatype)
+				.read();
+		InformationModel model = (InformationModel) workspace.get().stream().filter(p -> p instanceof InformationModel)
```
aedelmann: Hi @MayilKannanKumarBosch 
What does this test case test "testAppend" ?? I do not see any asserts .
this applies for all other tests you have written.

PR #1887, opened 2019-08-21
Junit tests

MayilKannanKumarBosch:


aedelmann:
@MayilKannanKumarBosch 
You have merge conflicts. Please do not use the same branch after you have created a PR. Otherwise you run into same issues and need to resolve conflicts because your old branch was branched off from an older commit id.

PR #1950, opened 2019-09-16
Mapping Pipeline Tutorial Update

tmtnckff:
Updated URLs, added description texts and images


tmtnckff:
@aedelmann please review

PR #1959, opened 2019-09-18
Feature/issue 1952

erlemantos:


erlemantos:
It is keeping in spirit with all our Events listener. All of our Events listener ends with Listener.

aedelmann:
How about CommentAnonymizationListener ?

Issue #1964, opened 2019-09-19
OpenAPI Generator - Missing resources


philbuettner:
The resources "/things", "/things/{thingId}", "/things/{thingId}/features/{featureId}/properties" and "/things/{thingId}/features/{featureId}/properties/{propertyPath}" are missing from the generated openAPI specifications. However they are available in the original IoT Things REST API. See https://apidocs.bosch-iot-suite.com/?urls.primaryName=Bosch%20IoT%20Things%20-%20API%20v2#/.

For the purpose of completion, these resources should also be available in the generated documentation.

Details:

- /things resource should be added to the generated openAPI specifications. The return type should use the generic "Things" schema defined by IoT Things REST API.

- /things/{thingId} resource should be added. The return type should use the specific schema defined by the corresponding vorto definition. For example, for smart home controller openAPI specification, the schema should be the JSON structure of SHC thing.
- /things/{thingId}/features/{featureId}/properties resource should be added. The return type should use the specific schema for the properties. For example, for feature DeviceInfo, the DeviceInfoProperties should be used.
- /things/{thingId}/features/{featureId}/properties/{propertyPath} should be added. The definition for the IoT Things REST API can be used completely.

PR #2007, opened 2019-10-09
#1964

philbuettner:
Added Things resource: /things &  /things/{thingId}

### DIFF
--- a/generators/generator-openapi/src/main/java/org/eclipse/vorto/codegen/openapi/templates/OpenAPITemplate.xtend
+++ b/generators/generator-openapi/src/main/java/org/eclipse/vorto/codegen/openapi/templates/OpenAPITemplate.xtend
@@ -55,6 +55,8 @@ class OpenAPITemplate implements IFileTemplate<InformationModel> {
 		  - url: https://things.eu-1.bosch-iot-suite.com/api/2
 		    description: "Bosch IoT Things Service"
 		tags:
+		  - name: Things
+		    description: List every Thing - «Utils.getReferencedFunctionBlocks(infomodel)»
 		  - name: Features
 		    description: Features of your «infomodel.name» things
 		  - name: Messages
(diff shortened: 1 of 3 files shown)


generators/generator-openapi/src/main/java/org/eclipse/vorto/codegen/openapi/templates/OpenAPITemplate.xtend
```diff
@@ -373,6 +482,30 @@ class OpenAPITemplate implements IFileTemplate<InformationModel> {
...
+		      - features
+		      properties:
+		        thingId:
+		          type: string
+		          description: Unique identifier representing the thing
+		        policyId:
+		          type: string
+		          description: The policy ID used for controlling access to this thing, managed by resource `/policies/{policyId}`
+		        attributes:
+		          $ref: '#/components/schemas/Attributes'
```
aedelmann: Where is attributes defined ?

generators/generator-openapi/src/main/java/org/eclipse/vorto/codegen/openapi/templates/OpenAPITemplate.xtend
```diff
@@ -373,6 +482,30 @@ class OpenAPITemplate implements IFileTemplate<InformationModel> {
...
+		        features:
+		          $ref: '#/components/schemas/Features'
```
aedelmann: Where is Features defined ? Shouldn't this be the «infomodel.name»Features ?

philbuettner: Attributes is defined in line 503.

generators/generator-openapi/src/main/java/org/eclipse/vorto/codegen/openapi/templates/OpenAPITemplate.xtend
philbuettner: Attributes is defined in line 503.

generators/generator-openapi/src/main/java/org/eclipse/vorto/codegen/openapi/templates/OpenAPITemplate.xtend
philbuettner: I will add FeatureDefinition and FeatureProperties in a next step. to be in line with the IoT Things schema.

aedelmann: Hi @philbuettner 
AFAIK the Features specifically for the information model are already generated. In this case the Thing model would directly reference that one. The one you said from existing Things Schema is a generic structure , and not strong typed to the function blocks.

philbuettner: Ok, what should be the content of /things/{thingId} then?

aedelmann: the same Thing model as you have specified except that the **features** references #schema/components/«infomodel.name»Features

aedelmann:
@philbuettner Some of your commits are not signed off. Can you do that, otherwise I cannot merge.

PR #2024, opened 2019-10-15
JUnit12

MayilKannanKumarBosch:
Signed-off-by: MayilKannanKumarBosch <MayilKannan.Kumar@bosch-si.com>


generators/generator-boschiotsuite/boschiotsuite-cloud/src/test/java/org/eclipse/vorto/codegen/bosch/templates/BoschIoTSuiteGeneratorTest.java
```diff
@@ -0,0 +1,79 @@
...
+  /*
+   * Check if SingleGenerationResult is returned when config property is passed as provision to
+   * BoschIOTSuiteGenerator
+   */
+  @Test
+  public void generateProvisionTrue() throws Exception {
+    configProperties.put("provision", "true");
+    InvocationContext context = new InvocationContext(mappingModels, configProperties);
+    assertEquals(true, boschIOTSuiteGenerator.generate(modelProvider(),
+        context) instanceof SingleGenerationResult);
```
aedelmann: Shouldn't it check if it returns a provisioning script file ?

PR #2025, opened 2019-10-15
Bug/issue 2016 docker images push fix

somesh-kumar:


somesh-kumar:
Change docker login username switch from -username to -u

PR #2056, opened 2019-11-01
JUnit20

MayilKannanKumarBosch:
Signed-off-by: MayilKannanKumarBosch <MayilKannan.Kumar@bosch-si.com>


aedelmann:
Why excluding it instead of writing unit tests ?

PR #2084, opened 2019-11-19
Make content in event and operation blocks required. Fixes #2058

### LINKED ISSUE #2058: Editor creates duplicates of fields
In some cases, the editor creates duplicates of the operation and event properties.

Exemplary, I have created the following functionblock:
```vortolang 1.0
namespace myNamespace
version 0.1.0
displayname "DO"
description "Digital Output"

functionblock DO {
    status {
        mandatory DO as boolean "if the output is active=true or inactive=false"
    }
    
    operations {
        on() "activates the output"
    }
    
    events {
    }
}
```

After saving, the model is modified to:

```
vortolang 1.0
namespace myNamespace
version 0.1.0
displayname "DO"
description "Digital Output"

functionblock DO {

    status {
        mandatory DO as boolean "if the output is active=true or inactive=false"
    }
    
    operations {
        on() "activates the output"
    }
    
    events {

    }

operations {
    }
    
    events {

    }

}
```

kolotu:
This makes at least one element in the events{} and operations{} blocks required. Users will still be able to load models that have been created with empty blocks before this change, however they cannot be saved with empty blocks again. 


aedelmann:
Build is failing due to some test case failure with your change. Can u take a look ?

kolotu:
Yes already on it. One of the test cases checks a model that's now invalid. I'll remove the empty operations block from the affected test case and re-run the test.

kolotu:
The second commit should fix the integration tests. I've also manually tested the OpenAPI, Ditto and Hono generators locally with a model containing empty events and operations blocks - the output looks good to me.

PR #2096, opened 2019-11-20
Feature/issue 2090 fail clm scan on error notify GitHub

somesh-kumar:


somesh-kumar:
@aedelmann please merge, getting rid of the variable since we can't use it in declarative pipeline without script. Cleaned it up.

PR #2118, opened 2019-11-28
Fixed #2063 aws stacktrace in cloudwatch and ram monitoring in linux …

### LINKED ISSUE #2063: AWS: Improve Error Alerts with Stack Trace
**Actual**: From the time we enabled Alerts for exceptions in production logs, we get a message with no mention of what really went wrong. The only way to know is by going into the logs.

**Expected**: Add a stack trace in the log, so a developer can decide if a fix is really required and decide upon the urgency of the fix.

somesh-kumar:
…ami's

Signed-off-by: Somesh Kumar <somesh.kumar@bosch-si.com>


somesh-kumar:
.ebextensions folder has been copied to s3

PR #2127, opened 2019-12-02
upgraded jgroups version for security vulnerability

aedelmann:


somesh-kumar:
Can't find a change log for 3.6.11, it was released in 2017, there is a 3.6.19 available, no clue what is fixed in there as well. How do we decide if 3.6.11 is good enough?

aedelmann:
Hi @somesh-kumar 
with 3.6.11 the CLM does not report the vulnerability anymore. I guess its the right version then :) However it seems that all warnings in CLM will also make the scan fail. Is this intentional ? There are 30+ warnings. We would fix all of them in order for this scan to succeed. I do not see this realistic.

PR #2144, opened 2019-12-05
enhancement/2132 Integration Tests for Elastic Search functionality i…

mena-bosch:
enhancement/2132 Integration Tests for Elastic Search functionality in Vorto Repository

This change set is mostly about adding integration tests to the Elasticsearch module:

1. Added over 100 integration tests mostly ported from the ones created for the repository-core, with necessary changes
2. Ported SearchTestInfrastructure from repository-core: it spins an Elasticsearch runtime with Allegro on ports differing than defaults to avoid conflicts - see class javadoc
3. Modified POM to accomodate the above and various dependencies (including XTend plugin to validate repository imports)
4. Added model resources for the tests (a subset of resources used in repository-core)
5. Removed some garbage comments in repository-server POM
6. Fixed some typos in repository-core tests where expected result was substituted with test result in assert by mistake

### DIFF
--- a/repository/repository-elasticsearch/pom.xml
+++ b/repository/repository-elasticsearch/pom.xml
@@ -1,32 +1,97 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-	<modelVersion>4.0.0</modelVersion>
-	<parent>
-		<groupId>org.eclipse.vorto</groupId>
-		<artifactId>repository</artifactId>
-		<version>1.0.0-SNAPSHOT</version>
-	</parent>
-	<artifactId>repository-elasticsearch</artifactId>
-	<packaging>jar</packaging>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.eclipse.vorto</groupId>
+    <artifactId>repository</artifactId>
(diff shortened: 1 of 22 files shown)


mena-bosch:
One thing worth noticing @aedelmann : [Allegro embedded Elasticsearch](https://github.com/allegro/embedded-elasticsearch) is covered by Apache 2.0. Do you know if there is any specific action I should take or just referencing the dependency as-is in the pom is sufficient in terms of compliance?

mena-bosch:
Looks like the new integration tests introduced by this PR are failing on the remote machine, but succeeding on my machine and another (Tobias'). 
This could be for the following reasons (in decreasing order of likeliness):

- Allegro is configured to spin Elasticsearch on ports 19200/1 and 19300/1. If the machine's firewall is configured to block those ports, that will be a problem
- Allegro needs to download an Elasticsearch runtime from the web before spinning it up. That again could be a problem if the machine is either firewalled or has restricted internet access.
- Allegro downloads the runtime and create temporary data in the /tmp directory. Read/write access and the actual existence of the directory are a requisite. 
- In one test class, the initial import of the model on BeforeClass requires reindexing. It turns out this _also_ requires some waiting time before starting the tests. This has been worked around by adding a very ugly Thread.sleep statement in the end of the BeforeClass method, in order to allow the Elasticsearch runtime to catch up before tests are performed. I could not find a less inelegant mechanism so far, and of course, the sleeping time is arbitrary and might not be sufficient on the remote machine / might be required for more/all test classes.

aedelmann:
Hi @mena-bosch 
I accidentally merged the PR already. but it would be good, if you could work on a fix to not let travis fail. In the meantime, until this is fixed , can you raise a PR excluding these tests for the time being. And then create a separate branch/ PR containing the fix? What do you think ?

mena-bosch:
Hi @aedelmann, I can probably butcher something to temporarily disable the tests. The _real_ fix should be on the machine itself. As we found out, it will likely have something to do with proxy configuration. Will get back to you later today on the temporary disabling.

mena-bosch:
Just to align here - I understand from this morning's call that the tests have already been disabled on the dev machine? I cannot see anything in source control, so wondering if I still need to do this while the guys figure out how to fix the proxy on the machine?

aedelmann:
Erle had disabled the tests for elastic search module in the Pom.xml.

Issue #2130, opened 2019-12-02
Repository OAuthProvider API


erlemantos:
**Why**
3rd party users of Vorto can use the Repository with their OAuth2 Provider containing users.
This requires an API which allows to plugin a Vorto OAuthProvider easily and configure the repository with it. There are no "core" or UI changes needed in order to use each provider.

**Proposal of the API**

```
IRepositoryOAuthProvider {
    String getID() //e.g. SiemensID
    boolean verify(IAuthentication token)
}
```

The Vorto Repository "ships" with implementations for this API, namely Github, BoschID, BoschIoTSuiteOAuth

PR #2146, opened 2019-12-06
Feature/issue 2130

erlemantos:
Fixes https://github.com/eclipse/vorto/issues/2130


repository/repository-core/src/main/java/org/eclipse/vorto/repository/sso/oauth/TokenVerificationProvider.java
```diff
@@ -13,21 +13,37 @@
 package org.eclipse.vorto.repository.sso.oauth;
 
 import javax.servlet.http.HttpServletRequest;
+import org.springframework.security.core.Authentication;
 import org.springframework.security.oauth2.provider.OAuth2Authentication;
 
 public interface TokenVerificationProvider {
```
aedelmann: Interfaces should start with an I according to the eclipse guidelines :)

PR #2176, opened 2019-12-13
Remove quotation marks from Content-Disposition header. Fixes #2168

### LINKED ISSUE #2168: Exception 500 when invoking generators on Vorto dev system
Steps taken:
- choose any Information model and go to its details page
- Generate code with any generator
Expected: Generated code can be downloaded successfully
Actual : 500 Internal Error

kolotu:
@aedelmann I have removed the quotation marks from the header (see explanation in my comment in #2168). Or do you see any reason, why this should not be done?


aedelmann:
No. This looks good to me. Thanks for the fix !

PR #2190, opened 2019-12-19
Moved definition field up to thing-level. Fixes #2188.

### LINKED ISSUE #2188: Move the "definition" of the generated Things structure up to thing-level
Currently the definition field that contains the Vorto model ID in the generated thing JSON is nested inside the attributes. Due to the newly-defined "definition" attribute on the thing-level, this has to be moved up.
Also the definition on the thing level is an array of string, so the Vorto model ID has to be wrapped in an array. 

Example for current generation:
```
{
    "thingId": "com.kevin.suite.tutorial:my-device-fromVortoPublic2",
    "acl": {},
    "attributes": {
      "thingName": "WasherWAT286H8SG",
      "definition": "org.eclipse.vorto.com.bshg:WasherWAT286H8SG:2.0.0"
    },
    "features": {
      "washer": {
        "definition": [
          "org.eclipse.vorto.com.bshg.programs:Washer:2.0.0"
        ],
        "properties": {
          "status": {
            "program": "Cotton",
            "temperature": "Cold",
            "spinSpeed": "Off",
            "delayedStart": 0,
            "delayedStartUnit": ""
          }
        }
      }
   }
}
```

Example of the required generation: 
```
{
    "thingId": "com.kevin.suite.tutorial:my-device-fromVortoPublic2",
    "definition": ["org.eclipse.vorto.com.bshg:WasherWAT286H8SG:2.0.0"],
    "acl": {},
    "attributes": {
      "thingName": "WasherWAT286H8SG"
    },
    "features": {
      "washer": {
        "definition": [
          "org.eclipse.vorto.com.bshg.programs:Washer:2.0.0"
        ],
        "properties": {
          "status": {
            "program": "Cotton",
            "temperature": "Cold",
            "spinSpeed": "Off",
            "delayedStart": 0,
            "delayedStartUnit": ""
          }
        }
      }
  }
}
```

kolotu:


generators/generator-eclipseditto/src/main/java/org/eclipse/vorto/codegen/ditto/schema/tasks/template/DittoStructureTemplate.xtend
```diff
@@ -30,13 +30,13 @@ class DittoStructureTemplate implements IFileTemplate<InformationModel> {
 	override getPath(InformationModel context) {
 		return null;
 	}
-	
+
 	override getContent(InformationModel model, InvocationContext context) {
 		'''
 {
+  "definition": ["«model.namespace»:«model.name»:«model.version»"],
```
thjaeckle: Hi @kolotu - this is not exactly correct.
The "definition" on Thing level is not a Json array, but a Json string, see also the documentation: https://www.eclipse.org/ditto/basic-thing.html#api-version-2

generators/generator-eclipseditto/src/main/java/org/eclipse/vorto/codegen/ditto/schema/tasks/template/DittoStructureTemplate.xtend
```diff
@@ -30,13 +30,13 @@ class DittoStructureTemplate implements IFileTemplate<InformationModel> {
...
   "attributes": {
- 	"thingName": "«model.displayname»",
- 	"definition": "«model.namespace»:«model.name»:«model.version»"
+ 	"thingName": "«model.displayname»"
```
thjaeckle: "thingName" could be confusing for the user as Ditto names the part of the thingId after the `:` also "thingName", at least in the model classes.
Maybe better use "displayName" or even "modelDisplayName".

generators/generator-eclipseditto/src/main/java/org/eclipse/vorto/codegen/ditto/schema/tasks/template/DittoStructureTemplate.xtend
kolotu: Hi @thjaeckle - thanks for pointing that out, I'll fix it right away in this issue: 
https://github.com/eclipse/vorto/issues/2213

generators/generator-eclipseditto/src/main/java/org/eclipse/vorto/codegen/ditto/schema/tasks/template/DittoStructureTemplate.xtend
kolotu: modelDisplayName does in fact describe this field more accurately, especially, since the value of this field will only be stored as additional attribute in Ditto. 
https://github.com/eclipse/vorto/issues/2215

PR #2214, opened 2020-01-10
Fixed the definition field. Changed the array to a string. Fixes #2213.

### LINKED ISSUE #2213: Fix Ditto Definition Field
The definition field in the Thing JSON must be a String instead of an array.

kolotu:


kolotu:
Fix included in PR #2216

Issue #2217, opened 2020-01-10
bug: concurrent running of ElasticSearch tests may produce interferences due to same ports in use


mena-bosch:
This is a speculative resolution for a problem we have noticed only recently. 

One (or more) of the integration tests for ElasticSearch was failing to reindex a model when during the build on the Jenkins side, showing a socket timeout exception. 

Given how those tests work (an ES runtime is launched by Allegro on specific ports), I suspect that multiple concurrent builds testing the ES integration tests at the same time might actually be stepping on each others' feet, when the build processes are not isolated. 

I will try and randomize the ports used to spin the service with the following rules:

* must be 4 digit ports
* must not be the production ES ports, or other obvious entries

Hopefully this will limit the collisions well enough for the use case.

mena-bosch:
PR [here](https://github.com/eclipse/vorto/pull/2218).
Local tests doing good - hopefully no surprise upstream.

mena-bosch:
Looks like this solved _one_ problem, aka the socket timeouts, but still does not allow multiple non-isolated builds running the ElasticSearch tests concurrently - maybe it's just a limitation on Allegro's side...
Working with @t-gauss to see if we can't force throttling builds. 

Not sure what could be an alternative solution should that fail, save for disabling those tests (which does not sound like a great idea).

t-gauss:
Hi @mena-bosch ,
being still a newbie in this topic I guess adding disableConcurrentBuilds() in the options block of the jenkinsfile could help as a temporary workaround.

t-gauss:
Hi @mena-bosch ,
did the change and ran the tests locally before commit. Even there I got a timeout 
`Tests in error: 
  org.eclipse.vorto.repository.search.TypeSearchSimpleTest: Failed to start elasticsearch within time-out
  org.eclipse.vorto.repository.search.TypeSearchSimpleTest
`
Will look into it.

t-gauss:
PR [here](https://github.com/eclipse/vorto/pull/2224).

mena-bosch:
Thanks @t-gauss . This one was merged by @kolotu now so closing.

Issue #2210, opened 2020-01-09
jQuery 3.2.1 has been found to be vulnerable to prototype pollution (CVE-2019-11358)


mena-bosch:
We are specifying the jQuery version explicitly in the repository pom. 

Looking at [this](https://docs.angularjs.org/misc/faq#does-angularjs-use-the-jquery-library-) page, AngularJS does not specify a compatibility grid with jQuery. 

However in [this](https://docs.angularjs.org/misc/version-support-status) page, AngularJS LTS versions are supposed to be maintained _also_ based on further jQuery releases that would break AngularJS. 

Under the latter assumption, I am going to try and update the referenced jQuery version to latest (3.4.1 at the time of writing). 

The vulnerability should be fixed starting from jQuery 3.4.0 (see [here](https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/)).

If upgrading jQuery alone doesn't break anything, it should resolve the issue. 
If it does break the front-end, we may need to consider updating AngularJS to latest (1.7.9) as well, and iterate.

mena-bosch:
PR [here](https://github.com/eclipse/vorto/pull/2211). This will fail as there is another compliance issue to handle, but hopefully whoever (probably myself) handles the other one can merge this branch first.

mena-bosch:
Thought this was merged already - reopening until it is.

kolotu:
PR merged.

mena-bosch:
Now merged for real.

PR #2220, opened 2020-01-13
2219 make 0.13 migration sql scripts remove redundant column

mena-bosch:
also merged latest jQuery and ElasticSearch tasks (#2210 and #2217) to avoid failing build


repository/pom.xml
```diff
@@ -255,7 +255,7 @@
 			<dependency>
 				<groupId>org.webjars</groupId>
 				<artifactId>jquery</artifactId>
-				<version>3.2.1</version>
```
aedelmann: Why u upgrading to lib in this PR. Need to create CQ at eclipse.

mena-bosch: @aedelmann I preemptively merged the branch where I updated the jquery version. 
The update to 3.4.1 comes from the fact that we had an advisory with an open vulnerability or prototype pollution (CVE-2019-11358) - see #2210. 
As the pull request for that was not merged yet, I have been merging the branch in all my recent PRs in order to remove some of the compliance issues.

mena-bosch: On second thought: 

1. The jquery version change is already merged in dev, because @kolotu merged my branch into one of his before PRing and his was merged. 
2. Paradoxally, the original PR for the security fix is still not merged.
3. Yes, given this is a intermediate version change required to address the CVE, a CQ should be started with Eclipse. I am not familiar how to do this as of yet, but I suspect it becomes necessary when releasing (i.e. not when just merging to dev branch). If you have the time, I think it would be worth having a look together (possibly with Kevin and Tobias as well), since we're going to need that anyway.

kolotu: @aedelmann @mena-bosch 
I have created the CQ - it has been approved.

PR #2222, opened 2020-01-13
Bug/2221 the angular controller for search does not tag model types

mena-bosch:
Quick and easy fix on controller to prepend type tag to relevant search when using drop-down.


mena-bosch:
The compliance issue with Spring seems to be back, despite having been waived (for non-usage) by @t-gauss . Not sure I understand how this works.

t-gauss:
I tried again with a scoped waiver. Hopefully its gone now.

PR #2224, opened 2020-01-14
added config option to disable concurrent builds. Fixes #2217

### LINKED ISSUE #2217: bug: concurrent running of ElasticSearch tests may produce interferences due to same ports in use

t-gauss:
Signed-off-by: Tobias Gauss <tobias.gauss@bosch-si.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,6 +3,7 @@ pipeline {
     // Options covers all other job properties or wrapper functions that apply to entire Pipeline.
     options {
         buildDiscarder(logRotator(numToKeepStr:'5'))
+        disableConcurrentBuilds()
     }
     stages{
       stage("Build"){


mena-bosch:
Here the ES tests failed, but in Travis (supposedly isolated) and due to a "different" timeout as discussed. 
Had not seen this one before - hopefully just a glitch. Will be restarting this soon.

mena-bosch:
ES tests are failing again due to inability to manipulate the temporary Allegro/ES data in the /tmp folder. 
Restarting everything as I am not sure whether this was cause by builds running concurrently.

mena-bosch:
Not sure whether this works yet. When I restart jenkins builds manually, they are running concurrently so that would mean the config change does not work in that context. However, there might be a chance for it to work exclusively when jenkins starts the jobs from new PR detected. 
Guess we'll just have to wait and see.

t-gauss:
Think it works only in scope of PR in jenkins. But lets try the measures we discussed.

Issue #2229, opened 2020-01-15
Unable to run ElasticSearch integration tests in parallel


mena-bosch:
This issue relates to a few issues already created, namely #2217 and #2224 - none of which solved the original problem so far. 

#2217 was about randomizing ports to avoid collisions between parallel builds when running Allegro+ES but it was not sufficient to prevent issues that ultimately failed the tests (e.g. later issues were about Allegro "unable" to access the /tmp folders it requires for temp data, broadly speaking). 

#2224 tackled the problem at Jenkins level, by attempting to throttle concurrent builds. Aside from the fact that it doesn't seem to work at least when builds are (re-)started manually, it seems a bit overkill to limit our build concurrency just for some integration tests to work. 

The current situation leaves 3 options open (in order of priority to investigate):

1. Investigate the Allegro framework to see if we didn't miss out an existing option to allow running it concurrently
2. If not, either consider getting rid of Allegro altogether and spinning a Docker container with ES directly (port randomization would then be very relevant)
3. Or, investigate the Allegro source itself and find a way to allow concurrent runs - then propose it in the project and use a fork in the meantime

Will start with n. 1 shortly.

mena-bosch:
Regarding point 1 - research what Allegro offers in terms of build/test parallelism. 

There is a [section](https://github.com/allegro/embedded-elasticsearch#running-more-then-one-elasticsearch-instance) of their documentation that specifically addresses this... 
... but fails to provide clarity on our specific use case. 

We already have `PopularProperties.TRANSPORT_TCP_PORT` and `PopularProperties.HTTP_PORT` randomized per implementation of #2217 , which did solve _one_ of the issues but not the overall problem. 

We do not use `withInstallationDirectory`, so that section of the docs is either inaccurate (cannot reuse same installation dir when using concurrently) or irrelevant to our case. 

Per @t-gauss ' changes, we are using `withInResourceLocation` pointing to a zip with ES. 

**Things to try**

- Specify a randomized `withInstallationDirectory` for each running test, **or**
- Not clean the installation directory (`withCleanInstallationDirectoryOnStop(false)` - defaults to `true` when not specified). 

These experiments will require:

1. Having two fake pull requests to run concurrent builds
2. Unblocking our pipeline (currently stuck at yet another compliance issue only @t-gauss can waive)

Will wait until tomorrow and get cracking again if all goes well.

mena-bosch:
[...]

PR #2234, opened 2020-01-16
#2229

mena-bosch:
* tried randomizing ES installation directory too, as a string composed of random ports and relative to running target

### DIFF
--- a/repository/repository-elasticsearch/src/test/java/org/eclipse/vorto/repository/search/SearchTestInfrastructure.java
+++ b/repository/repository-elasticsearch/src/test/java/org/eclipse/vorto/repository/search/SearchTestInfrastructure.java
@@ -15,6 +15,7 @@ import static org.mockito.Mockito.when;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


mena-bosch:
For clarity: this is an experiment to see if specifying the Allegro ES installation directory (in my case, a string composed of static chars and the random ports, in a path relative to the target folder of the build) helps with Allegro not breaking during concurrent builds, aka concurrent ES runtimes. 
Another thing to try within scope if this still creates issues is to disable Allegro from cleaning up after itself, since that folder will be cleaned up by maven anyway. 
Once merged, will test with two concurrent empty pull requests. 
Might need to revert if no benefit found.

mena-bosch:
I have tried testing this with two no-change pull requests 2229_test_0 and 2229_test_1. 
Unfortunately, it appears only the travis build is triggerd. 
Maybe the jenkins job requires actual change to trigger. 
Will try again with PRs with harmless changes (e.g. some javadoc). 
They are obviously not meant to be merged, but rather just triggers to see if the jenkins builds run concurrently without the ES tests breaking.

mena-bosch:
I stand corrected. 

2229_test_0 is **now** running the other checks including the jenkins build. 

2229_test_1 still is in an all-green state, but maybe this is just slow. 

Will wait it out a while longer.

mena-bosch:
Jenkins build for 2229_test_1 appears to have now started as well. 
Hopefully some clarity later this afternoon.

mena-bosch:
Looks like both builds succeeded, but I cannot be certain the ES integration tests really ran at the same time. 
For safety, I will need to re-run the jenkins builds manually once again. 
Also worth noting, this PR has a small bug: the ES install directory is set as a relative path to the module's home instead of the target folder. 
I already have a trivial fix waiting locally, so I will pull request it soon.

mena-bosch:
Fix is up and merged. Merging dev into both empty branches and restarting builds manually for both once build for extraneous PR 2241 is finished.

Issue #2221, opened 2020-01-13
Bug: the angular controller for search does not tag model types


mena-bosch:
This is a residual bug that was not one before the refactory of the search functionality planned for 0.13 (see #1833 and #2046). 
Prior to the new search, models were searchable by type without a tag. 

This has been removed in the new search to avoid confusion between type and name, i.e. only name can be searched untagged. 

However, I omitted the part of the Javascript controller that handles the types drop-down: it needs to prepend the "type:" tag to the search expression sent to the back-end. 

Otherwise, the type search would be interpreted as a name search, and since the untagged name search also covers the description, it would always yield all models for that type. 

The fix is trivial and just consists in pre-pending the type tag when using anything but "all types" in the relevant dropdown.

mena-bosch:
PR [here](https://github.com/eclipse/vorto/pull/2222). 
@aedelmann @kolotu Definitely needed for 0.13.

mena-bosch:
Restarted build ensuring single process in non-isolated CI, which passed all checks. @kolotu just merged, so closing this.

PR #2241, opened 2020-01-17
Bug/2221 the angular controller for search does not tag model types

mena-bosch:


mena-bosch:
Something went wrong with previous PR for #2221 and as a result the only relevant commit was not merged.

PR #2244, opened 2020-01-20
bug/2242 Set a longer timeout for ES integration tests using allegro

mena-bosch:
* set timeout to 1 minute (default was 15 seconds)


mena-bosch:
Sigh. 
Build passed, but we have new scanned vulnerabilities - one of which has been waived as non relevant in the past and is coming back in disguise. 
Will have to analyze the new one and ask @t-gauss to re-waive the first one.

PR #2246, opened 2020-01-21
bug/2243 Users without the Publish Permission on a namespace can see …

mena-bosch:
…the Publish button

* changed the pre-authorization of ModelRepositoryController#getPolicies to all users
* added call to above endpoint in detailsController to infer whether user can publish
* replaced permission check in template to use flag from controller rather than full access permission


mena-bosch:
> Looks good to me, since you filter the policies and only return those for the user.

And the user is inferred in the back-end, not given by the front-end, so there's no using this endpoint to check on permissions for other users either.

PR #2259, opened 2020-01-27
Prevent properties from being added multiple times. Fixes #2172

### LINKED ISSUE #2172: OpenAPI Generator - entities are generating duplicated mapping keys
When generating OpenAPI documents, in some documents the following error appears:
Parser error - duplicated mapping key

This error only appears in connection with an entity object inside of a .type model. Because the OpenAPI Generator Template function, which parses entities did not changed from current release (0.12.4) to development branch, the cause of that error must be somewhere else (in 0.12.4 this error does not appear).

This function:
https://github.com/eclipse/vorto/blob/2657685397bcf9a0f361843d8508fffcead3eef2/generators/generator-openapi/src/main/java/org/eclipse/vorto/codegen/openapi/templates/OpenAPITemplate.xtend#L615

generates this error in the development branch:
```
    Incident:
      type: object
      properties:
        incidentType:
          $ref: '#/components/schemas/IncidentType'
        roomId:
          type: string
        timeOfIncident:
          $ref: '#/components/schemas/DateTime'
~~~ Parser error - duplicated mapping key
        incidentType:
          $ref: '#/components/schemas/IncidentType'
        roomId:
          type: string
        timeOfIncident:
          $ref: '#/components/schemas/DateTime'
```

The following OpenAPI documents are containing the duplicated keys. If you compare them with each other, there are also slightly different in the block order. Documents which are generated correctly, always have the same order. But I am not sure if this was by coincident.

[OpenAPIgeneratedWithDuplicatedKeys.zip](https://github.com/eclipse/vorto/files/3951189/OpenAPIgeneratedWithDuplicatedKeys.zip)

t-gauss:
Signed-off-by: Tobias Gauss <tobias.gauss@bosch-si.com>

### DIFF
--- a/core-bundles/meta-model/org.eclipse.vorto.core/src/org/eclipse/vorto/core/api/model/ModelConversionUtils.java
+++ b/core-bundles/meta-model/org.eclipse.vorto.core/src/org/eclipse/vorto/core/api/model/ModelConversionUtils.java
@@ -236,6 +236,7 @@ public class ModelConversionUtils {
     newProperty.setPresence(property.getPresence());
     newProperty.setMultiplicity(property.isMultiplicity());
     newProperty.setDescription(property.getDescription());
+    newProperty.setConstraintRule(property.getConstraintRule());
 
     if (property.getType() instanceof PrimitivePropertyType) {
       PrimitivePropertyType newPrimitiveType =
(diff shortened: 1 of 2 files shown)


t-gauss:
Now the patterns are missing - need to re-check.

PR #2272, opened 2020-02-02
Fix typo in references to FontoXPath

DrRataplan:
FotonXPath would also be a cool name, but the project is called fontoxpath.

Thanks for finding and using FontoXPath though!


DrRataplan:
I have signed the ECA, but that does not seem to be automatically detected.

kolotu:
Hi @DrRataplan, 
Thank you for your contribution and sorry for that type in your project's name. 
Could you please verify that you have used the same email address for both the commit and the ECA? Also please make sure that all commits are signed (using git commit -s). I think that would be the reason why the check is failing.

DrRataplan:
@kolotu Thanks, that was it: I accidentally committed using my work e-mail address. The check is working now.

kolotu:
@DrRataplan cool thanks for updating the PR. I have merged it now.

PR #2300, opened 2020-02-19
Added guava version (18)

mena-bosch:
Modified poms accordingly to avoid using guava 22 snapshot (dead) indirectly

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,7 @@ pipeline {
             withMaven(
                 maven: 'maven-latest',
                 mavenLocalRepo: '.repository') {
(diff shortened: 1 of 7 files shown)


mena-bosch:
If travis build passes, this will have to be merged in currently failing PRs.

mena-bosch:
Travis build passes but jenkins builds to not. I can parametrize the Jenkins file to use -U - this should sort out the issue. 
Might need some trial and error to see if a one time -U, then revert is sufficient to get rid of the issue definitely, or whether we need to leave -U. 
Not very happy with this but it's blocking all further PRs...

mena-bosch:
Closing as this was not the suitable solution after all.

PR #2306, opened 2020-02-20
2299 update xtext

mena-bosch:


mena-bosch:
Compliance check on transitive dependency for xtend failed - suspect it was waived in the past because it consists in the tool not finding a license, whereas a license is indeed present i the jar. 
Reference: lsp4j 0.8.1 used by xtend 2.20. 
@t-gauss waived the compliance issue.

kolotu:
Merged after compliance issue was waived.

PR #2307, opened 2020-02-21
bug/2298 Display a validation error message in the technical user dia…

mena-bosch:
…logue

* Added pattern validation message for subject in the form


mena-bosch:
Merged this on my own as it's rather a trivial bit of UI code.

PR #2316, opened 2020-02-25
Added project version and revision number to UI footer.

kolotu:
And project version to generators


kolotu:
> Can't remember - don't we want to update the copyright year when performing changes on a resource?

yes - fixed it now.

generators/generator-boschiotsuite/boschiotsuite-cloud/src/main/java/org/eclipse/vorto/codegen/bosch/BoschIoTSuiteGenerator.java
```diff
@@ -1,13 +1,12 @@
...
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
- *
- * See the NOTICE file(s) distributed with this work for additional
- * information regarding copyright ownership.
- *
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License 2.0 which is available at
- * https://www.eclipse.org/legal/epl-2.0
- *
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
```
mena-bosch: Suspect this would fail the plugin when building due to paragraphs and line returns :(

kolotu: should be fixed now.

PR #2319, opened 2020-02-26
UX/2305 Cannot change image of my vorto model

mena-bosch:


mena-bosch:
Some tests I forgot about. Fixing shortly.

PR #2329, opened 2020-03-02
Bug/2326 uploading large images fails without producing a human readable error to the end user

mena-bosch:


repository/repository-core/src/main/java/org/eclipse/vorto/repository/web/core/ModelRepositoryController.java
```diff
@@ -176,27 +181,51 @@ public void getModelImage(
...
+      return new ResponseEntity<>(
+          AttachResult.fail(
+            actualModelID,
+            file.getOriginalFilename(),
+            String.format(
+              "The attachment is too large. Maximum size allowed is %dMB",
+              attachmentValidator.getMaxFileSizeSetting()
+            )
+          ),
+          HttpStatus.PAYLOAD_TOO_LARGE
```
kolotu: nice!

PR #2336, opened 2020-03-05
JCR Session handling optimization. Fixes #2320

### LINKED ISSUE #2320: Inefficient JcrSession handling
For a single requests (e.g. getModelConent) there are many jcr sessions opened and closed which causes high load on the DB. Session should be reused at least in http request scope.

In addition to that for each session which is opened there are additional requests to the DB to fetch the roles for the corresponding user.

t-gauss:

### DIFF
--- a/repository/repository-core/src/test/java/org/eclipse/vorto/repository/core/RequestRepositorySessionHelperTest.java
+++ b/repository/repository-core/src/test/java/org/eclipse/vorto/repository/core/RequestRepositorySessionHelperTest.java
@@ -16,27 +16,25 @@ import org.eclipse.vorto.repository.AbstractIntegrationTest;
 import org.eclipse.vorto.repository.core.impl.RequestRepositorySessionHelper;
 import org.junit.Assert;
 import org.junit.Test;
-import org.xml.sax.ContentHandler;
-import org.xml.sax.SAXException;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 
-import javax.jcr.*;
-import javax.jcr.lock.LockException;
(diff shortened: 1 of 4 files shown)


repository/repository-core/src/test/java/org/eclipse/vorto/repository/core/RequestRepositorySessionHelperTest.java
```diff
@@ -0,0 +1,347 @@
...
+        // new session should be returned
+        Session sessionForReuse = helper.getSession();
+        Assert.assertFalse(mySession.equals(sessionForReuse));
+        // after destroy session should still be alive
+        helper.destroy();
+        Assert.assertTrue(sessionForReuse.isLive());
+
+    }
+
+    private Repository createMockRepository() {
```
kolotu: Wouldn't it be easier to use Mockito here? Or is there a reason not to use it for this test?

t-gauss: Hi @kolotu ,
thanks. I will adapt this.

mena-bosch: Can't check out the branch right now but I agree with the mock-related comment. Also (and forgive me if that's obvious) - are the invocation to `get` on the supplier always guaranteed to not return `null`?

t-gauss: Hi @mena-bosch ,
I didn't change the behaviour compared to the way it was before.

t-gauss:
@kolotu Thanks. That looks much better readable.

PR #2347, opened 2020-03-17
Fixed elevated session handling

kolotu:
Signed-off-by: kolotu <kevin.olotu@bosch.io>


mena-bosch:
Might use some tests in later PR but I think it's more urgent to deploy the fix as of now.

PR #2352, opened 2020-03-19
bug/2349 Non-Systemadmin users can't save models

mena-bosch:
* fixed namespace/tenant equality comparison being too strict wrt to subdomains when checking for user access


mena-bosch:
Merging despite errors, as CLM scan looks broken (issues out of the blue with no library change, no access to abstract for anyone in the team).
Change set is minimal, fixes a regression in production and no change in libraries. 
Travis build fails due to one test which is known to fail intermittently (not impacted by current fix anyway).

Issue #2350, opened 2020-03-19
After restore of a namespace private models are visible to anon users


t-gauss:
If restoring a namespace (or the complete repository) the AccessControlList gets lost. As a result a non-logged in user can see private models.

t-gauss:
PR [here](https://github.com/eclipse/vorto/pull/2362)

PR #2356, opened 2020-03-23
Bug 2355 save models broken

mena-bosch:


mena-bosch:
Closing without merging. A new version of this temporary fix is available and it will go straight to prod as an interim solution, which should be fixed by #2350 ultimatly.

PR #2357, opened 2020-03-23
Bug 2355 save models broken

mena-bosch:
deploying this to prod for tomorrow, so we can replicate with additional server-side logs


mena-bosch:
CLM scan s still broken and no visibility on actual issue. Merging this to deployment as I have a session with impacted user tomorrow and this serves the purpose of augmenting the logs.

PR #2361, opened 2020-03-25
bug/2355_save_models_broken

mena-bosch:
* fixed saveModel in-body authorization for tests as at least one test impacted is using multiple namespaces per tenant


mena-bosch:
One test was failing as it was using multiple namespaces per "tenant". 
Will probably fail at CLM level again unless it's been fixed in the meantime - waiting it out and/or merging to deployment later.

PR #2364, opened 2020-03-26
Restore access policies mechanism and endpoint added. Fixes #2350

### LINKED ISSUE #2350: After restore of a namespace private models are visible to anon users

mena-bosch:


mena-bosch:
Merging to deployment too as this should be in prod asap.

PR #2369, opened 2020-03-27
bug/2368

mena-bosch:
* Fixed check for namespace quota, now including private namespaces


mena-bosch:
Build passed - merging as compliance checks are still broken.

PR #2372, opened 2020-03-31
2354_cors

mena-bosch:
* Added www.eclipse.org origin to patterns, as swagger requests somehow have that as origin


mena-bosch:
merging right away - small config change only and need to test with swagger asap

PR #2387, opened 2020-04-21
2380 suite auth access to public models

kolotu:
There was a check in the OAuth handler, that checked whether the OAuth client was a technical user in the namespace of the requested model. IMO this is not necessary as the regular access control applies. 
@mena-bosch maybe have a quick look, to see if you see a reason, why this maybe shouldn't be removed.


mena-bosch:
> There was a check in the OAuth handler, that checked whether the OAuth client was a technical user in the namespace of the requested model. IMO this is not necessary as the regular access control applies.
> @mena-bosch maybe have a quick look, to see if you see a reason, why this maybe shouldn't be removed.

In all honesty I have no idea what this business-level authorization is doing in one specific OAuth provider, as I would expect the services or controllers relevant to the request to handle this. 

I vote to remove.

kolotu:
Merging with compliance check failures as they are not caused by the content of this PR.

PR #2388, opened 2020-04-21
Fixes #2386

### LINKED ISSUE #2386: Local generator execution using SAM-cli always returns base64 encoded String
When you try to execute the generators locally using the sam cli, the generator result is always base64 encoded which is not very convenient (you need to manually decode it). The local execution and execution in AWS should behave the same.

t-gauss:


kolotu:
Merged with failed compliance check, as the failure does not affect the content of the PR.

PR #2391, opened 2020-04-21
Duplicated properties in generators. Function block hierarchy in Ditt…

### LINKED ISSUE #2374: Respect Functionblock Hierarchy in Ditto JSON Generator
Thanks @thjaeckle  for reporting this and providing an example:

> When using inheritance in Vorto Functionblocks, Ditto relies on that all Functionblocks in the hierarchy are added to the "definition" array, the first entry being the direct type.
> For example given that you have 3 functionblocks extending each other:

```
functionblock ColoredLamp extends Lamp {
...
}
functionblock Lamp extends Switchable {
...
}
functionblock Switchable {
...
}
```

> And a Ditto Feature "my-colored-lamp" shall follow the "ColoredLamp" definition.
> Then this must look like:

```
{
  "thingId": "foo:bar",
  "policyId": "foo:bar",
  "features": {
    "my-colored-lamp": {
      "definition": [
        "org.eclipse.vorto:ColoredLamp:1.0.0",
        "org.eclipse.vorto:Lamp:1.0.0",
        "org.eclipse.vorto:Switchable:1.0.0"
       ],
       "properties": {
          ... whatever ...
        }
     }
  }
}
```

> What is important is that "ColoredLamp" is the first entry (e.g. Developer Console relies on that in order to identify the actual followed type).
> 
> This "flattening" of the inheritance hierarchy is needed by Ditto because otherwise Ditto won't be able to search for all features which implement the "org.eclipse.vorto:Switchable:1.0.0" functionblock.
>

### LINKED ISSUE #2376: Generators - function block inheritance leads to multiplied properties
Became visible while working on #2374
Example:
Functionblock A extends Functionblock B extends Functionblock C

Expected result when generating the JSON:
Functionblock A should contain properties of Functionblock B and properties of Functionblock C

Actual result:
Functionblock A contains properties of Functionblock B and each inherited property from Functionblock C **two times**

Please assign the issue to me.

t-gauss:
…o generator. Fixes #2376, Fixes #2374


core-bundles/meta-model/org.eclipse.vorto.core/src/org/eclipse/vorto/core/api/model/ModelConversionUtils.java
```diff
@@ -104,6 +101,18 @@ public static FunctionblockModel convertToFlatHierarchy(FunctionblockModel fbm)
...
 
+  private static List<Property> removeDuplicates(List<Property> allProperties) {
+    return allProperties.stream()
+            .filter(filterByTypeAndName(p -> p.getName() + "_" + p.getType().getClass().getName()))
+            .collect(Collectors.toList());
+  }
+
+  public static <T> Predicate<T> filterByTypeAndName(
+          Function<? super T, ?> generateKey) {
+    Map<Object, Boolean> insertedPropertiesMap = new ConcurrentHashMap<>();
```
kolotu, line 112: Why did you use ConcurrentHashMap here?

generators/generator-eclipseditto/pom.xml
```diff
@@ -41,6 +41,12 @@
 			<version>4.12</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.skyscreamer</groupId>
```
kolotu, line 45: Is this a new dependency?

t-gauss: As far as I know we have it already as transitive dependency of serveral other libraries

kolotu: I couldn't find any use of it in the code and also no CQ - I will create a CQ for it.

core-bundles/meta-model/org.eclipse.vorto.core/src/org/eclipse/vorto/core/api/model/ModelConversionUtils.java
t-gauss: Because non concurrent map would not work with parallel streams. But you are right- where its used now the stream is sequential. Do you want me to replace it with a non concurrent map?

kolotu: I think a non-concurrent HashMap would be more efficient in terms of performance.

kolotu:
Merged with failures in the compliance check, as the content of the PR is not affected by it.

PR #2403, opened 2020-04-29
license plugin configuration adapted and file headers added. Fixes #2317

### LINKED ISSUE #2317: License Check Plugin does not scan the generators
The classes in the generators module also need to be checked for the right license headers.
Xtend, Xtext and JS files also need to be included in the scan.

t-gauss:


kolotu:
Merged with failures in the compliancy check, as they are not caused by the contents of this PR.

PR #2406, opened 2020-05-04
Add Function Block Mapping to Spec-TOC.

JulianFeinauer:
Signed-off-by: julian <j.feinauer@pragmaticminds.de>


kolotu:
Merged with failures in checks, since the failures are not caused by the content of this PR.

PR #2407, opened 2020-05-04
update sonarcloud token

kolotu:


kolotu:
Merged without checks, because it's a one-line change in the Jenkinsfile.

PR #2408, opened 2020-05-04
re-added PR config

kolotu:


kolotu:
merged without checks, since it is a change in the build pipeline.

PR #2409, opened 2020-05-04
UX/2351 After manipulating users on namespaces, the namespace filter …

mena-bosch:
…resets while still displayin...


mena-bosch:
Merging regardless of compliance failures - those are already being handled in separate tickets and the CQs are still awaiting approval.

PR #2412, opened 2020-05-06
2411/Create a local mysql profile to troubleshoot persistence with th…

mena-bosch:
Added the profile and reference in ES config. 


mena-bosch:
Merging right away as this is just a new profile.

PR #2417, opened 2020-05-18
2291

mena-bosch:
Added feature to request access to namespaces. 
When the namespace has administrators who configured their user with an e-mail address, it will send a message to those administrators. 
Will require configuring the dev and ultimately prod environments with the appropriate SMTP props. 
This is loosely based on the Vorto e-mail templating and sending capabilities in-app, which are however never actually used.


mena-bosch:
Merging to dev - nothing can be done about compliance checks until CQs are approved and they have been in limbo for weeks now. 
Need to test the e-mail sending on dev.

PR #2427, opened 2020-05-26
2426

mena-bosch:
Upgrades AngularJS to latest
Removes direct references to jQuery in POM
Adapts notation of REST calls in UI controllers
Adapts $location notation everywhere


mena-bosch:
Merging to dev regardless of new CVE on latest AngularJS. 
We need to test this on dev, and the new CVE can be analyzed and hopefully waived (no upgrade path) - at worst it can wait until we rewrite the UI with a non-obsoleting framework. 
The advantage is that we get rid of the CVE this task aimed at resolving, plus the ones on jQuery (and we can withdraw the CQ on jQuery, in favor of a new one on latest AngularJS).

PR #2430, opened 2020-05-27
2426

mena-bosch:
re-added jQuery (3.5.1, was 3.4.1) as it seems relying on jQLite alone doesn't render the UI


mena-bosch:
Merging right away, will have to CQ jQuery 3.5.1

PR #2433, opened 2020-05-28
2432 Upgrade jQuery to latest version

mena-bosch:

### DIFF
--- a/repository/pom.xml
+++ b/repository/pom.xml
@@ -269,7 +269,7 @@
 			<dependency>
 				<groupId>org.webjars</groupId>
 				<artifactId>jquery</artifactId>
-				<version>3.4.1</version>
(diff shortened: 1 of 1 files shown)


mena-bosch:
Ignoring security compliance issues with Tomcat (will deal with later), Spring security core (idem), AngularJS 1.7.9 (has been waived with valid justifications but system did "not pick it up"). 
Latest compliance issue with jQuery 3.5.1 seems to be related to a "license not found" issue, rather than a CVE. 
As there's no known literature wrt license changes between jQuery 3.4.1 and 3.5.1 we can assume this is an issue on Sonatype, and waive accordingly. 
Merging to dev - will create CQ on jQuery 3.5.1 shortly.

PR #2440, opened 2020-06-04
2438

mena-bosch:
Fixed self-closing span into explicitly closed to not render button caption as part of span with wrongly inherited font-family


mena-bosch:
Compliance check unrelated and will be addressed in other task.

PR #2449, opened 2020-06-15
Added documentation for adding oauth clients

kolotu:


kolotu:
Merged without running the checks, because the changeset only consists of documentation files.

PR #2452, opened 2020-06-16
2450 improve documentation

kolotu:


kolotu:
Merged as compliance checks are not caused by contents of the PR.

PR #2453, opened 2020-06-16
2425

mena-bosch:


mena-bosch:
Ignoring AngularJS CVEs as there is no upgrade path. 
Spring core CVE to be handled in separate task.

PR #2455, opened 2020-06-17
2448/Messages after saving a model are not always displayed

mena-bosch:
Fixes some bad references to data coming from http calls after upgrade to AngularJS 1.7.9 and related syntax changes


mena-bosch:
Pushing without review as I must have accidentally formatted the resource. Unlikely to break more than it fixes. Will iterate with team when available on dev.

PR #2458, opened 2020-06-18
2457/Redirect requests to new AngularJS notation

mena-bosch:
* removes bang from location hash so AngularJS URLs are backwards-compatible after framework upgrade
* re-fixes all links again


mena-bosch:
Ignoring failures at this time - some infra issues and the compliance items are handled somewhere else.

Issue #2447, opened 2020-06-10
Improving the vortolang spec for measurementUnit


kolotu:
From the specification it's hard to understand the correct syntax for the use of 'measurementUnit' (see here: https://github.com/eclipse/vorto/blob/development/docs/vortolang-1.0.md)

The documentation should be clarified and maybe provide an example to show, how the syntax works.

PR #2467, opened 2020-06-27
Issue #2447 Add example how to use measurement unit

JulianFeinauer:
Signed-off-by: Julian Feinauer <j.feinauer@pragmaticminds.de>


kolotu:
@JulianFeinauer thanks a lot for your contribution. What do you think about removing the readability from the example? It is redundant in the example, as status properties are readable by default and it makes the example a bit harder to read imho.

JulianFeinauer:
Agree, I removed it.

kolotu:
Merged without complete build, as this PR only contains changes in the docs.

PR #2515, opened 2020-08-13
Fix query param handling during redirect

kolotu:


kolotu:
forgot to add the cookie removal

PR #2533, opened 2020-08-17
Fixes #2532.

### LINKED ISSUE #2532: Deleting a namespace can break the repository
When attempting to delete a namespace as namespace admin, the modeshape workspace will be deleted first, user associations to the namespace and the entry in the namespace table will be deleted consecutively. 

If an error occurs while deleting user associations, the modeshape workspace is already deleted, but the namespace table entry remains, which prevents the repository from accessing any data in the entire modeshape repository.

A bug in the implementation prohibited any namespace admin from deleting user associations, so this would always break the repository

kolotu:

### DIFF
--- a/repository/repository-core/src/main/java/org/eclipse/vorto/repository/services/NamespaceService.java
+++ b/repository/repository-core/src/main/java/org/eclipse/vorto/repository/services/NamespaceService.java
@@ -13,13 +13,6 @@
 package org.eclipse.vorto.repository.services;
 
 import com.google.common.collect.Lists;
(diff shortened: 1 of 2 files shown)


kolotu:
This change deletes the user associations and the namespace table entry before deleting the modeshape workspace. It also fixes a check that validates whether a namespace admin has the permission to remove the association from a user to a namespace.

kolotu:
Changed the order of deletion to: user associations, modeshape workspace, namespace table as the modeshape deletion method accesses the namespace table.

PR #2553, opened 2020-08-20
Fixes #2552.

### LINKED ISSUE #2552: Loading model details also triggers diagnostics for that model if the user is sysadmin
As discovered recently, loading a model's details when the user has the sysadmin repository role _also_ triggers a diagnostic for that model, which slows down the loading. 

The bit of code doing this has been unchanged for a while now, and there is no clear reason why loading a model as sysadmin should trigger those diagnostics (the functionality is available for the whole lot in the manage section). 

While this would not improve the performance for most Vorto users, it would for vorto developers who happen to be sysadmins. 

The change set is a two liner delete in the details AngularJS controller.  

Low prio but easy fix.

kolotu:

### DIFF
--- a/repository/repository-web/dist/js/controllers/detailsController.js
+++ b/repository/repository-web/dist/js/controllers/detailsController.js
@@ -1025,17 +1025,6 @@ define(["../init/appController"], function (repositoryControllers) {
             );
           };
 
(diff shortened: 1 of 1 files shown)


kolotu:
Merged without travis build, since it's a minor UI change.

PR #2564, opened 2020-08-25
2562

mena-bosch:
Now with latest dev

### DIFF
--- a/repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/impl/ModelRepositoryFactory.java
+++ b/repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/impl/ModelRepositoryFactory.java
@@ -12,14 +12,42 @@
  */
 package org.eclipse.vorto.repository.core.impl;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.jcr.LoginException;
+import javax.jcr.Repository;
+import javax.jcr.RepositoryException;
+import javax.jcr.Session;
 import org.eclipse.vorto.model.ModelId;
-import org.eclipse.vorto.repository.core.*;
+import org.eclipse.vorto.repository.core.FatalModelRepositoryException;
+import org.eclipse.vorto.repository.core.IDiagnostics;
+import org.eclipse.vorto.repository.core.IModelPolicyManager;
+import org.eclipse.vorto.repository.core.IModelRepository;
+import org.eclipse.vorto.repository.core.IModelRepositoryFactory;
+import org.eclipse.vorto.repository.core.IModelRetrievalService;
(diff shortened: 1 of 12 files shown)


repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/impl/ModelRepositoryFactory.java
```diff
@@ -273,23 +297,50 @@ public IModelRepository getRepositoryByModel(ModelId modelId, IUserContext userC
...
-        userRoles.add(RepositoryRole.SYS_ADMIN);
-      }
-      return userRoles;
-    } catch (DoesNotExistException e) {
-      LOGGER.debug("User or namespace not found. ", e);
-      return Collections.emptySet();
-    }
+    String compositeID = workspaceId.concat(username);
+
+    Set<IRole> result = userNamespaceRolesCache.get(compositeID)
```
kolotu, line 321: I would probably place this logic inside the UserNamespaceRoleService, so it would be automatically used, wherever namespace roles are required.

kolotu (commented):
I'd place the cache lookup inside the getRolesByWorkspaceIdAndUser() method in the UserNamespaceRoleService, but apart from that it looks good to me.

repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/impl/ModelRepositoryFactory.java
mena-bosch: > I would probably place this logic inside the UserNamespaceRoleService, so it would be automatically used, wherever namespace roles are required.

It does make sense to broaden it, but given the context (request-scoped caching) I suspect there might not be many more scenarios availing of the cache.

kolotu: Fair enough - I couldn't tell if there are many more scenarios where this is helpful - we can also keep it as is and measure the impact first

t-gauss: The results for the specific test looked very good! I still think we should consider to implement a proper repository caching since this would improve the performance in general (given that there is a reasonable eviction strategy).

mena-bosch: > Fair enough - I couldn't tell if there are many more scenarios where this is helpful - we can also keep it as is and measure the impact first

I still think broadening per your suggestion is a good idea, if anything, for the "form" of it. Since this will have to wait a day or two, we have room to experiment.

mena-bosch: > The results for the specific test looked very good! I still think we should consider to implement a proper repository caching since this would improve the performance in general (given that there is a reasonable eviction strategy).

Sounds great. As we know, the caching may be tricky to implement given the situation but ultimately it is the Spring out-of-the-box tool for this matter, so it will likely be better than any homemade solution.

t-gauss (approved):

kolotu (approved):

PR #2631, opened 2020-10-12
2459

mena-bosch:
- Disabled forced-reloading of model after saving (likely unnecessary)
- Added info message from back-end in template when model saved successfully

### DIFF
--- a/repository/repository-web/dist/js/controllers/detailsController.js
+++ b/repository/repository-web/dist/js/controllers/detailsController.js
@@ -131,9 +131,7 @@ define(["../init/appController"], function (repositoryControllers) {
                 function (result) {
                   $scope.isLoading = false;
                   $scope.message = result.data.message;
-                  if (result.data.valid) {
-                    $scope.loadDetails();
-                  } else {
+                  if (!result.data.valid) {
                     $scope.validationIssues = result.data.validationIssues;
                   }
                 },
--- a/repository/repository-web/dist/js/controllers/detailsController.js
+++ b/repository/repository-web/dist/js/controllers/detailsController.js
@@ -131,7 +131,9 @@ define(["../init/appController"], function (repositoryControllers) {
                 function (result) {
                   $scope.isLoading = false;
                   $scope.message = result.data.message;
-                  if (!result.data.valid) {
+                  if (result.data.valid) {
+                    $scope.loadDetails();
+                  } else {
                     $scope.validationIssues = result.data.validationIssues;
                   }
                 },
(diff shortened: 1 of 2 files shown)


repository/repository-web/dist/js/controllers/detailsController.js
```diff
@@ -131,9 +131,7 @@ define(["../init/appController"], function (repositoryControllers) {
                 function (result) {
                   $scope.isLoading = false;
                   $scope.message = result.data.message;
-                  if (result.data.valid) {
-                    $scope.loadDetails();
```
kolotu, line 135: What was this originally supposed to do? Reload the details to update the last author/change date?

t-gauss: and the properties... e.g. the description.
reload

t-gauss (commented):
As mentioned I would suggest to add the "loadDetails()" again to avoid that the ModelInfo goes out of sync with the ModelEditor.

repository/repository-web/dist/js/controllers/detailsController.js
mena-bosch: I think it was "something like that". Basically reloading the whole of the model once the "Save" button was pressed, in order to refresh the data. 
So far, I haven't seen any of this requiring reloading the model though, so I figured I'd save us a round-trip to the back-end. 
Other functionalities requiring updating fields, such as releasing the model, etc. do not seem to be impacted.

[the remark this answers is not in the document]
mena-bosch: I have tried figuring out whether there was any difference before removing this, but couldn't find anything relevant. 
Do you have a concrete example?
Note that model comments have little/no applicability here since saving a comment does not go through this execution path.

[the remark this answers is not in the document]
mena-bosch: As discussed, the description and last modified by fields do indeed not update when saving here, so I might as well just re-enable a full reload.

kolotu (approved):

t-gauss (approved):

mena-bosch:
As discussed, temporarily restored the "loadDetails" invocation on save since some fields are not updated, which escaped me in the first tests. 
Also considering replacing this again with a simple assignment of the model in scope UI-side with the property of the payload returned by the PUT endpoint which should in theory represent the updated resource - will need to look around first and verify that the endpoint is implemented correctly in that regards.

mena-bosch:
> As discussed, temporarily restored the "loadDetails" invocation on save since some fields are not updated, which escaped me in the first tests.
> Also considering replacing this again with a simple assignment of the model in scope UI-side with the property of the payload returned by the PUT endpoint which should in theory represent the updated resource - will need to look around first and verify that the endpoint is implemented correctly in that regards.

Nope, that doesn't seem to work out of the box - likely the payload for the model embedded in the PUT response differs enough from what is returned by loadDetails that it isn't reusable as is. 
Better just reload the details at this point.

kolotu (approved):

PR #2648, opened 2020-10-16
Fixes #2629.

### LINKED ISSUE #2629: Check caching mechanism for Bosch IoT Suite generator postman/dev-console option
The Caching mechanism can pickup the wrong attachment in some cases. 

/api/v1/generators/boschiotsuite/models/org.eclipse.vorto.tutorials:RaspberryPi:1.0.0?provision=true&requestBodyOnly=true

for this generator it's <boschiotsuite> <true> <true> for the provisioning string
for the postman collection it's <boschiotsuite> <true> 
 
so it seems to get confused with the two tags with the same name here.

We could add another query parameter to the postman option (output=postman) or ensure that it's evaluated that there are either 1 or 2 <true> tags. 

This bug could also be already fixed and the data on this specific model was just outdated.

kolotu:

### DIFF
--- a/repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/IModelRepository.java
+++ b/repository/repository-core/src/main/java/org/eclipse/vorto/repository/core/IModelRepository.java
@@ -295,7 +295,7 @@ public interface IModelRepository {
    * @param tags
    * @return
    */
-  List<Attachment> getAttachmentsByTags(ModelId modelId, Set<Tag> tags)
+  List<Attachment> getAttachmentsByTags(ModelId modelId, List<Tag> tags)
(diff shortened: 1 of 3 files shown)


mena-bosch (commented):
Would it be worth it / feasible to add an integration test to cache, and then verify the cached value is consistent? Asking just in case - not too knowledgeable about the area.

kolotu:
@mena-bosch having integration test coverage for generators would definitely be nice to have, but since they are running outside of the repository, there is none currently. 
Setting this up will be a bigger task, I created a separate task here https://github.com/eclipse/vorto/issues/2649

mena-bosch (approved):

PR #2650, opened 2020-10-16
Sonarcloud travis

kolotu:
This addition to the Travis pipeline would update the base branch on Sonarcloud. Test coverage is not evaluated for some reason and the Travis job is marked as timeout failure, even though the command seems to be completed


kolotu:
https://travis-ci.com/github/bosch-io/vorto/jobs/400362698

PR #2655, opened 2020-10-20
2654

mena-bosch:
- Improved namespace validation/consistency both on the UI and back-end
- Better handling of namespace uppercase vs lowercase by lowercasing upon persistence and resolving case-insensitive
- Added coverage at integration and unit level for namespace format and resolution
- Added documentation on the above


kolotu (approved):
Code looks good, test cases provide the right coverage, docs also look good

t-gauss (approved):

Issue #2491, opened 2020-07-14
Default constraint does not work for float and boolean


kolotu:
```
status {
        abc as boolean <DEFAULT true>
}
```
Leads to error: 
Line 10 - This constraint value must be of a Short

mena-bosch:
Could not replicate for float. A DEFAULT constraint of either an integer or a dot-notation floating value works. 
@kolotu am I missing something crucial?

Does replicate with boolean, no boolean or actually no short value works there.

mena-bosch:
Also, boolean constraint issue replicates within a datatype declaration (not surprisingly since the constraint is defined in the datatype resource and used in the fb one).

kolotu:
@mena-bosch it does replicate for floats (and doubles) that end with .0 - i.e. 1.0
Seems like I only tested with .0 values before...

mena-bosch:
> @mena-bosch it does replicate for floats (and doubles) that end with .0 - i.e. 1.0
> Seems like I only tested with .0 values before...

It gets weirder. 

- `foo as float <DEFAULT 0.0>` --> ok
- `foo as float <DEFAULT 0.1>` --> ok
- `foo as float <DEFAULT 1.0>` --> `no viable alternative at input '1.0'`  ¯\_(ツ)_/¯
- `foo as float <DEFAULT 2.0>` --> ok again...

Also worth noting the parsing error messages are different in case of boolean vs float. 

- `boo as boolean <DEFAULT true>` --> `This constraint value must be of a Short `

Probably different issues in the grammar.

mena-bosch:
New and probably useless finding: the generated validation Java source does compare boolean representations with their values **enclosed in double-quotes**, so that never seems to work within context. 
It also sends the wrong error message (for the short type) when invalid, which "explains" the absurd validation error for the boolean constraint, but doesn't really add much to clarity.

mena-bosch:
> New and probably useless finding: the generated validation Java source does compare boolean representations with their values **enclosed in double-quotes**, so that never seems to work within context.
> It also sends the wrong error message (for the short type) when invalid, which "explains" the absurd validation error for the boolean constraint, but doesn't really add much to clarity.

Which leads me to the `AccordinglyValueValidator.xtend` resource. 
Can assume the "Long type" error message is a typo. 
Trying to loosen the comparison to at least add non-quoted "false" and "true" when evaluating boolean values.

mena-bosch:
[...]

PR #2657, opened 2020-10-22
2491

mena-bosch:
This tentative fix resolves the unparseability (unlexability?) of `1.0` float values e.g. in constraints, due to shadowing by the VortolangVersion enum. 

See #2491 (the part on float types only) for further details. 

The change comes with pros and cons. 

## pros

- Makes parsing float constraints with value `1.0` possible
- Still permits abstracting from the `1.0` default vortolang version, allowing further versions in the future

## cons

- The vortolang version is parsed with a float rule, which - despite appearances - is completely the wrong data type for a software version (however, the meta-model still specifies it is a string, so no repercussions in the java source)
- The original enum-based solution allowed an in-grammar validation of the `1.0` value. This is not possible now, which means an additional software layer is required to verify that version upon parsing, which adds some undesirable indirection to vortolang.


mena-bosch:
Looks like somehow the parent build doesn't build the metamodel before the language module?

aedelmann:
Hi @mena-bosch 
been a while since I had done vortolang.
I know you can use the character ^ , in order to escape a value which conflicts with a keyword in your grammar. So perhaps whenever somebody wants to use a float constraint of 1.0 , the person could write ^1.0. This of course requires the user to know that, but your editor could assist the user, by adding it automatically. 

Have you checked with the  xText team to find a good solution to this problem ?

mena-bosch:
Looks like somehow the parent build doesn't build the metamodel before the language module?

> Hi @mena-bosch
> been a while since I had done vortolang.
> I know you can use the character ^ , in order to escape a value which conflicts with a keyword in your grammar. So perhaps whenever somebody wants to use a float constraint of 1.0 , the person could write ^1.0. This of course requires the user to know that, but your editor could assist the user, by adding it automatically.
> 
> Have you checked with the xText team to find a good solution to this problem ?

Hi Alex, thanks for your reply. 
If I understand you correctly, you are suggesting changing the syntax for constraints by requiring a `^` (or whatever arbitrary non-word character) before floats? 
That would actually change the syntactical requirements of Vortolang - not sure I like the idea.

I'd be interested to convey the issue to the Xtext team as you recommend, but I don't know their requirements and timelines. 
For instance, I'm not sure how easy for me it would be to come up with a MCVE instead of just dumping the whole Vortolang to them and state my problem... do you have any advice that could help me get their attention?

Thanks again

core-bundles/language/org.eclipse.vorto.editor.datatype/src/org/eclipse/vorto/editor/datatype/Datatype.xtext
```diff
@@ -27,12 +28,16 @@ QualifiedName:
...
+  '1.0'
+;
+
 Type:
 	Entity | Enum
 ;
 
 Entity:
-   'vortolang' lang=VortoLangVersion
+   'vortolang' lang=FLOAT
```
kolotu, line 40: Why is the vortolang definition in the entity different from the one in the enum (line 56)?

core-bundles/language/org.eclipse.vorto.editor.datatype/src/org/eclipse/vorto/editor/datatype/Datatype.xtext
```diff
@@ -27,12 +28,16 @@ QualifiedName:
 	ID ('.' ID)*
 ;
 
+VortolangVersion:
```
kolotu, line 31: Am I reading it wrong that VortolangVersion is only used in the enum and otherwise it's defined as FLOAT? Shouldn't this be removed and also declared as FLOAT in the enum?

core-bundles/meta-model/org.eclipse.vorto.core/src/org/eclipse/vorto/core/api/model/BuilderUtils.java
```diff
@@ -126,9 +126,7 @@ public ModelBuilder(T model) {
     }
 
     public ModelBuilder<T> withVortolang(String lang) {
-      if ("1.0".equals(lang)) {
```
kolotu, line 129: Not sure, but since 1.0 is removed from the grammar validation, doesn't this check become more relevant now?

core-bundles/language/org.eclipse.vorto.editor.datatype/src/org/eclipse/vorto/editor/datatype/Datatype.xtext
mena-bosch: That is one of the reasons why the build isn't working - the other one being the rule is still there in Datatype.xtext :)
Looks like the merge from dev was more messed up than I thought. 
Am fixing locally and will push again.

core-bundles/meta-model/org.eclipse.vorto.core/src/org/eclipse/vorto/core/api/model/BuilderUtils.java
mena-bosch: It should not be there and there's another merge issue. pfffff
Thanks for noticing it.

core-bundles/language/org.eclipse.vorto.editor.datatype/src/org/eclipse/vorto/editor/datatype/Datatype.xtext
mena-bosch: Build is green now. Really not sure what happened SCM-wise but the right change set is now in after my last push.

kolotu:
The escaping character ^ did not work at the specific location before the DEFAULT value out of box, so I'm merging this grammar change.

mena-bosch:
> The escaping character ^ did not work at the specific location before the DEFAULT value out of box, so I'm merging this grammar change.

Thanks. Let's leave it on dev for a while and keep an eye on. 
If no better alternative comes up, then we can go with this one.

PR #2663, opened 2020-10-23
2660 fix docker image

kolotu:


kolotu:
This prevents the repository from failing to start. SuiteAuth login is not configurable, but it is arguable how configurable the Bosch ID option is in there anyway. Github authentication should work fine.

mena-bosch (approved):

PR #2669, opened 2020-10-27
Uitest save model notification

t-gauss:


mena-bosch (approved):
More elegant than what I had in mind / on my local branch - I'll reuse this thanks.

PR #2671, opened 2020-10-27
2670

mena-bosch:
- Reinforced model parsing by preventing unrecoverable NPE when model fields parsed as null (only seems to occur with UI / REST operations - integration tests work slightly differently)

- Added broad coverage for completely broken syntax, but no check on new error messages since integration test models are parsed differently - this will get better coverage at UI level


kolotu (approved):
looks good!

PR #2672, opened 2020-10-27
Modeshape doctor

kolotu:


repository/repository-web/dist/js/controllers/modeshapeDoctorController.js
```diff
@@ -0,0 +1,178 @@
...
+                $scope.modeshapePath = "";
+                $scope.modeshapeWorkspaceId = "";
+
+                $scope.readModeshapeData = function (modeshapeWorkspaceId, modeshapePath) {
+                    $http.get("/rest/namespaces/diagnostics/modeshape/node/" + modeshapeWorkspaceId + "?path=" + modeshapePath)
+                        .then(response => {
+                            $scope.modeshapeData = response.data;
+                            $scope.modeshapePath = modeshapePath;
+                        }).catch(error => {
+                        alert("Error reading modeshape node");
```
mena-bosch, line 31: I think JS alerts are a bit too rough for validation, or anything for that matter. As an alternative, I would propose either "toast-like" warnings, or "validation-like" warning (essentially warnings et al. handled in dynamic divs). There are a few usable examples across the code base.

repository/repository-web/dist/partials/admin/modeshapeDoctor.html
```diff
@@ -0,0 +1,112 @@
...
+    <h2 style="color: red"><i class="fa fa-exclamation-triangle"></i>Warning! Use the write/delete functions with caution!<i class="fa fa-exclamation-triangle"></i></h2>
+    <h4>Known limitations:</h4>
+    <ul>
+        <li>Writing values on multi-value properties is not supported.</li>
+    </ul>
+</div>
+<table class="table table-condensed">
+    <tr>
+        <td class="col-md-3"><label>Workspace ID:</label></td>
+        <td class="col-md-7"><input type="text" ng-model="modeshapeWorkspaceId" style="width: 100%"></td>
```
mena-bosch, line 12: Would it be possible to have some search functionality by namespace as an alternative? It's true that the ModeShape doctor is inherently for sysadmins who are expected to be able to query the namespace table for that workspace ID, but it would be much nicer IMO to have that straight away.

repository/repository-web/dist/partials/admin/modeshapeDoctor.html
```diff
@@ -0,0 +1,112 @@
...
+    </tr>
+    <tr>
+        <td class="col-md-3"><label>Modeshape Path: </label></td>
+        <td class="col-md-7"><input type="text" ng-model="modeshapePath" style="width: 100%"></td>
```
mena-bosch, line 16: Getting very confused with the appropriate path to use here. For instance if I set a breakpoint in ModelRepository#getById and try to use the string representation of a JCR node initialized there for a given model, it does not seem to work with the form... which brings me to my next point: could we have a small .md file linked in the UI with examples on which values are expected, how to look them up etc.?

mena-bosch: Another idea (keeping in mind this might require considerable extra work) would be to have an "auto-discover" feature where once the right workspace ID (or better, **namespace**) is punched in, the UI discovers the node hierarchy for you and exposes it through some tree widget.

mena-bosch: You could then select the node you want to work with from there.

kolotu: > Getting very confused with the appropriate path to use here. For instance if I set a breakpoint in ModelRepository#getById and try to use the string representation of a JCR node initialized there for a given model, it does not seem to work with the form... which brings me to my next point: could we have a small .md file linked in the UI with examples on which values are expected, how to look them up etc.?

The path that needs to be used is the standard modeshape path notation (with slashes) - only requirement is that the path is terminated with / as well. I'll add a hint on the UI.

mena-bosch (approved):

PR #2674, opened 2020-10-29
UI test enrichment

mena-bosch:
Adds a few tests loosely related to the model details UI, and a couple of tools.


mena-bosch:
After some discussion, I re-formatted the change set according to the intellij variant of the Google style guides. 
This means some changes are still unreadable, but limits the damage. 
Further on, we've agreed to not reformat non-compliant, existing resources but only the actual changes / new resources.

PR #2686, opened 2020-11-03
commons-io upgrade. Fixes #2684

### LINKED ISSUE #2684: Upgrade commons-io
Needs to be upgraded due to known security issue.

t-gauss:


mena-bosch (commented):
Wouldn't it be worth generalizing the version in the parent pom? 
This way, if (**when**) we need to update again, it's only in one place?
No comment my side on the version change itself as long as it compiles, runs and tests are green.

t-gauss:
Good point. We had two different versions in the past but since we have only one version now we can do as you suggested. I will update the PR tomorrow

t-gauss:
incorporated your review comments

mena-bosch (approved):

PR #2692, opened 2020-11-05
UITest stability fix

t-gauss:


mena-bosch (commented):
Did you figure out why looking elements by xpath is more reliable than by name? On paper, it looks like there shouldn't be any confusion in that DOM right? Just baffled here...

t-gauss:
Not sure. I'm not even sure if it was really the wrong element identified by findElementByName. Maybe somehow the state of the element was not correct (but I would expect an error message from selenium in this case). I really don't know - adding the click helped for me and it works on Jenkins as well.

mena-bosch:
> Not sure. I'm not even sure if it was really the wrong element identified by findElementByName. Maybe somehow the state of the element was not correct (but I would expect an error message from selenium in this case). I really don't know - adding the click helped for me and it works on Jenkins as well.

Fair enough, let's call it Selenium sorcery ^^

PR #2728, opened 2020-12-04
Fix/UI test stability fix

t-gauss:


repository/repository-server/src/test/java/org/eclipse/vorto/repository/server/ui/ModelDetailsUITest.java
```diff
@@ -63,15 +67,15 @@ public void testSaveModelSuccessfulNotification() {
...
     createModel().succeed();
     RemoteWebDriver driver = seleniumVortoHelper.getRemoteWebDriver();
-    // going full screen
-    driver.manage().window().fullscreen();
-    WebElement textArea = driver.findElementByXPath("//div[@class='ace_content']");
-    driver.executeScript("arguments[0].scrollIntoView();", textArea);
+    //find the save button
+    WebElement saveButton = driver.findElementByXPath("//a[@ng-click='saveModel()']");
+    // scroll down to the text area
+    driver.executeScript("window.scrollBy(0,350)");
```
ghost, line 73: I don't want to mess with this since it's finally stable now, but have you considered using `scrollIntoView` and parametrizing with the target element rather than use fixed pixel figures?

ghost (approved):

repository/repository-server/src/test/java/org/eclipse/vorto/repository/server/ui/ModelDetailsUITest.java
t-gauss: Yes. I have seen this approach. Unfortunately that was one of the instabilities. The referenced element was sometimes "detached from the dom" or in another state which caused the test to fail. I would suggest to leave it now as it seems stable at least on 3 different systems :-)

PR #2733, opened 2020-12-15
2529

ghost:
# Summary

- Migrates user and comment table to define identity and authorship as a combination of username + authentication provider "ID"
- Migrate codebase to force user resolution by the two criteria, as opposed to username only
- Implies new non-v1 REST signatures and relevant changes in UI, including standalone form parametrization
- Implies user search now identifying and disambiguating users with identical username, by OAuth provider
- Fixes in tests
- Some small extraneous bug fixes
- Some application aspects are not impacted by this, e.g. model authorship (both presentational and persisted)


kolotu (approved):

ghost:
Quick test done by @kolotu on migration scripts (building, then running repository locally twice in a row). 
This was a useful augmentation of the automatic and manual tests I performed before creating the PR, as the tables could have been "dirty" from too much manipulation in my local environment. 
Merging now to dev but as discussed, not deploying to production.

